Issue #19, opened 2015-10-20
Use SwiftLint


yoichitgy:
Add `.swiftlint.yml` and use SwiftLint.

https://github.com/realm/SwiftLint

yoichitgy:
Added by pull request #22.

PR #22, opened 2015-11-02
Added SwiftLint

yoichitgy:
Added SwiftLint (Issue #19).


yoichitgy:
Hound passes even in case an expression violating a rule of SwiftLint exists. Have to investigate more.

yoichitgy:
Confirmed Hound CI detected a SwiftLint violation. SwiftLint v0.4.0 was used.

PR #34, opened 2015-12-07
Adding property injection to containers

mowens:
Adding ability to load properties from resources (JSON, Plist) into a container which provides the ability to inject property values into service definitions. See Documentation/Properties.md for details


SwinjectTests/PlistPropertyLoaderSpec.swift
```diff
@@ -0,0 +1,22 @@
...
+
+class PlistPropertyLoaderSpec: QuickSpec {
+    override func spec() {
+        it("can handle missing resource") {
+            let loader = PlistPropertyLoader(bundle: NSBundle(forClass: self.dynamicType.self), name: "noexist")
+            let props = loader.load()
+            expect(props).to(beNil())
+        }
+    }
+}
```
mowens: Fixed in comment: 7cc6262cc1bba7b9be0c65317095f5032a4dc33c

yoichitgy:
Hi Mike. Thank you for adding the nice feature👍

I'll merge your PR. Before merging it, I'm trying two things:

1. Investigate whether `property<Property>(name: String, _ type: Property.Type) -> Property?` really needs the explicit type parameter.
2. Introduce a protocol to have `property` functions and to remove them from `Resolvable` protocol because they are only used in `factory` and `initCompleted` closures. If we name it `PropertyRetrievable`, the factory type is going to be `protocol<Resolvable, PropertyRetrievable> -> Service` or we can make a protocol inheriting both the protocols to replace `protocol<Resolvable, PropertyRetrievable>`. With this change, `SynchronizedResolver` does not need `property` functions.

What do you think about my ideas?

yoichitgy:
About the idea No. 1 above, I think we only need a function returning `Property?` without the property type parameter.

    public func property<Property>(name: String) -> Property? {
        return properties[name] as? Property
    }

Then change the unit tests to like this:

    properties.intValue = r.property("test.int")!
    properties.optionalIntValue = r.property("test.int")
    properties.implicitIntValue = r.property("test.int")

We get all the tests pass with the modifications.

This approach looks having another advantage in addition to the removal of the explicit type parameter. It is that users can see, by writing `!` by themselves,`property` method might fail. This is the same approach, as `resolve` method of `Container` or Swift-native `Dictionary` take, that `nil` is returned if the collection does not have the key (or name). If we force unwrap inside `property` method, we hide the possibility of crash from the users.

What do you think about the approach?

I would like to have your valuable input🚀

mowens:
@yoichitgy 

re: 1. Removing explicit type property function
That was my first attempt on the function and I could have sworn that optional Strings were not casting properly since the `Property` type was `Optional<String>` and the retrieved type from the map was an `NSTaggedPointerString`. If the single function does in fact work for all types then I am in favor to use the that. I will update things on my side later today and push those changes to this PR.

In regards to the force unwrap that might crash for a user. I think this is OK as really it is a developer mistake if they are trying to inject a property that doesn't exist into a required property. If they remove the property and don't update the signature of the Class/Struct/Property that is receiving the injection then their code is now broken. So IMO better to crash there. Either way you will crash somewhere since you have to unwrap the return value :)

re: 2. Moving property retrieve to its own protocol
+1 Definitely a better approach to separate these. I like the name `PropertyRetrievable` as well. I will make this change later today and push those changes to this PR

Another change that I wanted to make that I thought about last night was that the `protocol PropertyLoadable` really should be named `protocol PropertyLoaderType` to match Swift naming conventions. I will make this change as well if you aren't against it. Thoughts?

yoichitgy:
**Re: 1.** Thank you for your offer to update it to the single `property` function returning `Property?`.

Regarding the crash strategy, I agree with you. We have almost the same idea. A small difference is that the top of stack trace displayed on crash, whether Swinject framework code or user's code. I think crashing at user's code is less surprising to the user. Also it is an excuse as a framework author telling "you have to fix your code not to crash" ;)

**Re: 2.** Thank you for your update for this later😀

yoichitgy:
I'm completely positive to change `PropertyLoadable` to `PropertyLoaderType` to get better consistency.

yoichitgy:
(You've already pushed the update. Let me see the update.)

yoichitgy:
Thank you for update the code with the additional ones (like `ResolverType` in Container.Arguments.erb and code documentation). The way you fix is great, especially type aliasing `protocol<Resolvable, PropertyRetrievable>` to `ResolverType` with the good name, which I didn't aware when I wrote my feedback.

One last thing before I merge the PR (sorry I didn't mention before), I would like to change `load()` function to throwable one: `func load() throws -> [String : AnyObject]` because the possibility of crash is visible to the users. As we discussed in the different case above, it's good to crash the program if the property loader fails. The programmers should fix the crash before releasing their apps. The reason I prefer throwing an error to `fatalError` (if possible) is the uses can be aware the crash possibility by writing `try!`.

The error cases would be something like below (or anything with better names):

    public enum PropertyLoaderError: ErrorType {
        case InvalidJSONType(bundle: NSBundle, name: String)
        case InvalidJSONFile(bundle: NSBundle, name: String)
        case InvalidPlistFormat()
        // More cases if needed.
    }
    
    extension PropertyLoaderError: CustomStringConvertible {
        switch self {
            case . InvalidJSONType(let bundle, let name):
                return "Unable to cast JSON properties from bundle: \(bundle) for name: \(name)"
            case . InvalidJSONFile(let bundle, let name):
                return "Unable to read JSON properties from bundle: \(bundle) for name: \(name)"
            case . InvalidJSONType():
                return "Plist must be a top level dictionary"
            // More cases if needed.
        }
    }

Do you mind making `load()` throwing the error?

mowens:
@yoichitgy 
Done. `PropertyLoaderType` will now throw sane errors that bubble up to the user via `applyPropertyLoader`

yoichitgy:
Thanks @mowens for accepting my opinion! I'll merge the PR then release a new version tomorrow. Just sleeping time for me today.

Issue #38, opened 2015-12-09
Reduce the number of arguments on overloads of register and resolve methods


yoichitgy:
Currently 12 arguments are supported by overloads of `register` and `resolve` methods. They look overwhelming when displayed in an autocompletion popup of Xcode.

![screenshot](https://cloud.githubusercontent.com/assets/965994/11684237/8f9358b4-9eb3-11e5-9045-649946cf8092.jpg)

The number of supported arguments should be reduced before releasing version 1.0.0. Currently I am thinking to reduce them to **9 arguments** because "9" looks enough and not too short, and is 1-digit number. In total, 12 lines (3 arguments * 2 register/resolve methods * 2 with/without name parameter) can be removed from the popup.

This breaking change, though it has less impact, should be made before releasing version 1.0.0.

yoichitgy:
Fixed by PR #42.

Issue #35, opened 2015-12-08
Proposal for modular assembly of services in container


mowens:
# Proposal for modular assembly of services in container
Provide a simple mechanism to populate service definitions in a container that allows implementations to group related components together across frameworks/application

The current implementation has no opinion to how services are registered to the container which allows users to assembly their container(s) however you want. This is all well and good, but it would be nice to provide consumers of the framework an opinionated way to register their services. My proposal is to provide an `AssemblyType` and `Assembler` to provide such a feature. This approach would allow sane bootstrapping of the container (this concept is taking from is very vaguely taken from [Typhoon](https://github.com/appsquickly/Typhoon/wiki/Modularizing-Assemblies)

## Proposed Implementation
There are 2 components to the implementation: `AssemblyType` and `Assembler`. The `AssemblyType` provides an interface that can be implemented by classes/structs (depending on your use case) that is provided a container from elsewhere.

```
protocol AssemblyType {
    func assemble(container: Container)
}
```

The `Assembler` will take a list of `AssemblyType` instances and build the container. The `Assembler` will either:
a) Own the container
b) Create the container to be owned else where

The implementation would be really straight forward:

```
class Assembler {
   let container = Container()
   init(assemblies: [AssemblyType]) {
       for assembly in assemblies {
           assembly.assemble(container)
       }
   }
}
```

With this approach, one can separate their service definitions like so:

```
class ServiceAssembly: AssemblyType {
    func assemble(container: Container) {
        container.register(FooServiceType.self) { r in 
             FooService() 
        }
        container.register(BarServiceType.self) { r in 
            BarService() 
        }
    }
}

class ManagerAssembly: AssemblyType {
    func assemble(container: Container) {
        container.register(FooManagerType.self) { r in 
            FooManager(service: r.resolve(FooServiceType.self)!)
        }
        container.register(BarManagerType.self) { r in 
             BarManager(service: r.resolve(BarServiceType.self)!) 
        }
    }
}
```
Now like service definitions are grouped and we can build our container like so:
```
let assembler = Assembler(assemblies: [
    ServiceAssembly(),
    ManagerAssembly()
])
let container = assembler.container // will contain all registered services
```

Additionally, this provides an easy means to override service definitions with mocks using inheritance. Let's say we want to replace `FooService` with `MockFooService`:

```
class MockServiceAssembly: ServiceAssembly {
    override func assemble(container: Container) {
        super.assemble(container)
        
        container.register(FooServiceType.self) { r in 
[...]

PR #39, opened 2015-12-09
Adding Assembler feature for modular components

mowens:
Implementation for Issue #35


yoichitgy:
So quick action! I'll check the PR tomorrow. Sometimes Hound works incorrectly. You can simply ignore the warnings if you think Hound is wrong. Then I'll investigate Hound things later.

mowens:
@yoichitgy The 2 about the line endings are correct (doh!). I already fixed :+1:

Documentation/Assembler.md
```diff
@@ -0,0 +1,148 @@
...
+            }
+        }
+    }
+
+Here we have created 2 assemblies: 1 for services and 1 for managers. As you can see the `ManagerAssembly`
+leverages service definitions registered in the `ServiceAssembly`. Using this pattern the `ManagerAssembly`
+doesn't care where the `FooServiceType` and `BarServiceType` are registered, it just requires them to
+be registered else where.
+
+### AssemblyLoadAwareType
```
mowens: @yoichitgy Maybe it would be better to name this ```LoadAwareAssemblyType```?

yoichitgy:
@mowens 👍

yoichitgy:
I updated `.travis.yml` to finish running in shorter time (about 2-3 min shorter for each). You'll be able to see the Travis update if you rebase to `master` before you push the next time.

SwinjectTests/BasicAssembly.swift
```diff
@@ -0,0 +1,47 @@
...
+    }
+}
+
+class PropertyAsssembly: AssemblyType {
+    func assemble(container: Container) {
+        container.register(AnimalType.self) { r in
+            return Cat(name: r.property("test.string")!)
+        }
+    }
+}
```
mowens: @yoichitgy unsure how to fix these. The ```.swiftlint.yml``` is already ignoring the test directory and I can run swinftlint locally just fine against 0.4.0

Documentation/Assembler.md
yoichitgy: `XxxxAssemblyType` looks better, but how about using protocol extension like this?

    public protocol AssemblyType {
        func assemble(container: Container)
        func loaded(resolver: ResolverType)
    }

    extension AssemblyType {
        func loaded(resolver: ResolverType) {
        }
    }

SwinjectTests/BasicAssembly.swift
yoichitgy: Hound looks ignoring `excluded` settings in `.swiftlint.yml`. I excluded test code because unit tests using Quick usually get very long.

You can ignore the Hound warnings for unit tests though `BasicAssembly.swift` has no new line at end and `LoadAwareAssembly.swift` has two new lines at end. (Just small things. You can fix the two files if you care.)

Swinject/Container.swift
```diff
@@ -100,7 +100,7 @@ public final class Container {
     ///
     /// - parameter loader: the loader to load properties into the container
     ///
-    public func applyPropertyLoader<T: PropertyLoaderType>(loader: T) throws {
+    public func applyPropertyLoader(loader: PropertyLoaderType) throws {
```
yoichitgy: Did you just make simpler here? I want to learn the intention here to write Swift code😀

Swinject/Assembler.swift
```diff
@@ -0,0 +1,99 @@
...
+    /// lazy load several assemblies that have interdependencies between each other use `appyAssemblies`
+    ///
+    /// - parameter assembly: the assembly to apply to the container
+    ///
+    public func applyAssembly(assembly: AssemblyType) {
+        assembly.assemble(container)
+        if let assemblyLoadAware = assembly as? AssemblyLoadAwareType {
+            assemblyLoadAware.loaded(resolver)
+        }
+    }
```
yoichitgy: I noticed the case `assemblyLoadAware.loaded(resolver)` is called is not covered by any test.

screenshot

Maybe this method can be implemented as below to simplify?

    public func applyAssembly(assembly: AssemblyType) {
        applyAssemblies([assembly])
    }

SwinjectTests/AssemblerSpec.swift
```diff
@@ -0,0 +1,232 @@
...
+    override func spec() {
+        
+        describe("Assembler basic init") {
+            it("can assembly a single container") {
+                let assembler = try! Assembler(assemblies: [
+                    AnimalAssembly()
+                ])
+                let cat = assembler.resolver.resolve(AnimalType.self)
+                expect(cat).toNot(beNil())
+                expect(cat!.name) == "Whiskers"
```
yoichitgy: 🐱

SwinjectTests/AssemblerSpec.swift
```diff
@@ -0,0 +1,232 @@
...
+                expect(sushi).toNot(beNil())
+            }
+        }
+        
+        describe("Assembler load aware") {
+            it("can assembly a single container") {
+                let loadAwareAssembly = LoadAwareAssembly() { resolver in
+                    let cat = resolver.resolve(AnimalType.self)
+                    expect(cat).toNot(beNil())
+                    expect(cat!.name) == "Bojangles"
```
yoichitgy: 🍗

yoichitgy:
Your code is clean and it was easy to read👍 I had good surprise that the implementation of the feature was so small! I wrote my feedback to the source code.

Since I updated some documents (`.md` files) after you make the branch, I would like you to rebase to fix the conflict. I just organized the documents, so I think the conflict is not so difficult.

Documentation/Assembler.md
mowens: Yea I can make this change. I thought it would be better to be explicit about which assemblies are going to run code after they are loaded by looking at its protocol name, but I don't ming the no-op function.

Swinject/Container.swift
mowens: In the `Assembler` you can pass an array of property loaders that will be applied, in order, to the container before the assemblies are loaded. Since the signature of the ```init``` function for the ```Assembler``` cannot constrain the array of property loaders generically it has be be passed as so:
```
init(...., propertyLoaders: [PropertyLoaderType], ...)
```
This created a problem for ```Container:applyPropertyLoader``` because the value passed for the ```loader``` param can only be constrained as a ```PropertyLoaderType``` not the type itself. So I had to make this change which doesn't break the API.

Swinject/Assembler.swift
mowens: :+1:

mowens:
@yoichitgy Thanks for the feedback and for the nice comments on my coding style :smile:

 I have incorporated your suggestions and have rebased master in my branch so the documentation is now current

Swinject/Container.swift
yoichitgy: Thank you for explaining the part. I got better understanding about the Swift generics.

yoichitgy:
@mowens Thanks for accepting my suggestions and rebasing to `master` for the documents😄

I'll fix Issue #38, then release a new beta version.

Issue #48, opened 2016-01-07
Swift Package Manager


ysholqamy:
Hi, are you willing to add support for swift package manager anytime soon?

yoichitgy:
Yes. I will support it soon. So far, I finished half things to do in my branch. Still I have to migrate unit test code to support it.

https://github.com/yoichitgy/Swinject/tree/source-layout

yoichitgy:
It looks if *SwinjectStoryboard* directory is in *Sources* directory, *Sources* directory cannot have source files right under the directory. I tried moving those files to *Sources/Swinject* directory, but still I got a lot of errors.

Need more investigation...

yoichitgy:
Will be available at v2.0.0-beta.2.

PR #54, opened 2016-01-21
Support Swift Package Manager

yoichitgy:
Modified the source layout of Swinject project to support Swift Package Manager (Issue #48).

Major changes are:

1. Use "Sources" and "Tests" directories.
2. Put files related to `SwinjectStoryboard` to "SwinjectStoryboard" directory.
3. Unify some files for iOS/tvOS and OSX with `#if os()"` because Swift does not allow having files with the same name even if they are in different directories.
4. Add an SPM manifest file.

Currently `SwinjectStoryboard` is not supported by SPM because it does not support Objective-C files at the moment.

`swift build` command reports a warning telling *"(FunctionType.swift.o) has no symbols"*, but it can be safely ignored:
https://bugs.swift.org/browse/SR-24


yoichitgy:
The Hound warnings can be ignored.

PR #58, opened 2016-02-01
Fixed assembler constructor: parent assembler is optional

Nikita2k:
Consider code:

```
// MyFile.swift

func getMyAssembler(parentAssembler: Assembler?) -> Assembler {
   if let parent = parentAssembler {
        return Assembler(parentAssembler: parent)
   } else {
        return Assembler() 
   }
}
```

The goal of PR is to simplify this function to: 
```
func getMyAssembler(parentAssembler: Assembler?) -> Assembler {
     return Assembler(parentAssembler: parent)
}
```

This is the PR I've mentioned here: https://github.com/Nikita2k/SwiftViper/issues/1

### DIFF
--- a/Sources/Assembler.swift
+++ b/Sources/Assembler.swift
@@ -30,8 +30,8 @@ public class Assembler {
     ///
     /// - parameter parentAssembler: the baseline assembler
     ///
-    public init(parentAssembler: Assembler) {
-        container = Container(parent: parentAssembler.container)
+    public init(parentAssembler: Assembler?) {
+        container = Container(parent: parentAssembler?.container)
     }
     
     /// Will create a new `Assembler` with the given `AssemblyType` instances to build a `Container`
@@ -56,8 +56,8 @@ public class Assembler {
     /// - parameter parentAssembler:    the baseline assembler
     /// - parameter propertyLoaders:    a list of property loaders to apply to the container
     ///
-    public init(assemblies: [AssemblyType], parentAssembler: Assembler, propertyLoaders: [PropertyLoaderType]? = nil) throws {
-        container = Container(parent: parentAssembler.container)
+    public init(assemblies: [AssemblyType], parentAssembler: Assembler?, propertyLoaders: [PropertyLoaderType]? = nil) throws {
+        container = Container(parent: parentAssembler?.container)
         if let propertyLoaders = propertyLoaders {
             for propertyLoader in propertyLoaders {
                 try self.container.applyPropertyLoader(propertyLoader)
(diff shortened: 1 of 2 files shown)


yoichitgy:
@Nikita2k thanks for sending a PR again😃, but I afraid the PR adds a bit complexity to the framework because `Assembler(parentAssembler: nil)` and `Assembler()` work identically. Users might be confused to choose an initializer if they don't use a parent assembler.

Alternatively, you can implement `getMyAssembler` function without merging the PR like this:

```swift
func getMyAssembler(parentAssembler: Assembler?) -> Assembler {
     return parentAssembler.map { Assembler(parentAssembler: $0) } ?? Assembler()
}
```

What do you think about my concern?

yoichitgy:
In the same way, my design of `Container.init(parent: Container? = nil)` might be a mistake.

Nikita2k:
In my opinion the most correct approach will be introducing `convenience init()` that will call `init(container: Container?)` and pass nil as a container.

Do you think this is better solution?

yoichitgy:
I think I rarely see `init()` as a convenience initializer. I would like to take time to investigate the best or a good practice of initializer design to avoid frequent breaking changes when we release new versions.

Have you ever seen any articles about the Swift initializer design?

Related topic:
A subclass of `Assembler` cannot use a convenience initializer in its initializer, though I guess `Assembler` should be `final class`.

yoichitgy:
So far, I couldn't find a good practice for the many initializers, but I've found this:

> Classes tend to have very few designated initializers, and it is quite common for a class to have only one. Designated initializers are “funnel” points through which initialization takes place, and through which the initialization process continues up the superclass chain.

> Convenience initializers are secondary, supporting initializers for a class. You can define a convenience initializer to call a designated initializer from the same class as the convenience initializer with some of **the designated initializer’s parameters set to default values**.

https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/Initialization.html#//apple_ref/doc/uid/TP40014097-CH18-ID203

Your idea `convenience init()` looks good.

Nikita2k:
@yoichitgy I'll change this PR to have convenience init. What do you think?

yoichitgy:
Let me take time a bit more? because it's going to be a breaking change.

yoichitgy:
@Nikita2k, I'm going to merge this pull request and release v1.1.0 because I want to align the assembler feature and #59 to release v2.0.0 with breaking changes. `convenience init()` and `final class Assembler` things are breaking changes strictly speaking, but actually less impact to existing users (they don't have to modify their code).

Are you ok with the merge of this pull request to release v1.1.0?

Nikita2k:
@yoichitgy sounds great!

yoichitgy:
@Nikita2k, thanks for the quick response😃

Nikita2k:
will fix convenience init soon

yoichitgy:
Thank you! We don't have to hurry. After resolving #59 is ok.

PR #78, opened 2016-04-05
Update Storyboard+Swizzling.swift

fractalzombie:
I think this will be better, don't like warnings.


yoichitgy:
Thanks @fractalzombie for the PR😃, but unfortunately Swift 2.1 doesn't support `#if swift(>= x.x)`. It can be used to distinguish Swift 3.0 from Swift 2.2 in the future as described in the note below.

https://swift.org/blog/swift-2-2-new-features/#compile-time-swift-version-checks

To confirm it, I tried to build the code using `#if swift` and couldn't compile it for an error.

yoichitgy:
The selector warnings were fixed by #89.

PR #109, opened 2016-06-10
Replace respondsToSelector call with optional chaining.

amccarri:
Found this while removing warnings from our project that uses Swinject.


yoichitgy:
Thank you for introducing the nice technique!

PR #116, opened 2016-06-25
Refactor the SpinLock class to use NSLock.

noremac:
http://engineering.postmates.com/Spinlocks-Considered-Harmful-On-iOS/

Based on this article, I've replaced `SpinLock`s usage of `OSSpinLock` with `NSLock`.


yoichitgy:
Thanks @noremac for the refactoring😃 ReactiveCocoa looked also changing SpinLock to another lock.

PR #134, opened 2016-08-25
Fix singleton creation

### LINKED ISSUE #133: ObjectScope .Container not working when using property injection
The following code will instantiate `A` twice even though it's supposed to be a Singleton.
```Swift
class A {
    init(b: B) {
        print("class A instatiated")
    }
}

class B {
    weak var a: A?
    init() {
        print("class B instatiated")
    }
}

let container = Container()

container.register(A.self) {
    r in
    A(b: r.resolve(B.self)!)
}.inObjectScope(.Container)

container.register(B.self) {
    r in
    B()
}.initCompleted {
    r, t in
    t.a = r.resolve(A.self)!
}.inObjectScope(.Container)

container.resolve(A.self)!
container.resolve(B.self)!
```
Prints:
```
class B instatiated
class A instatiated
class A instatiated
```

marcorei:
fixes #133 

Hello Yoichi,
thank you for creating Swinject!

I found that the calls to the `initCompleted` closures caused some objects to be created twice. In my case it was a ViewController which was instantiated twice during object creation.

Postponing the call to the `initComplete` methods allows the `Parent` class to be registered in the pool before the `initComplete` method of the `Child` gets executed and thus prevents the duplicate creation of `Parent`.

Please find in the PR a test case for the problem as well as a suggestion for the solution. This PR is targeted at version 1 since this is the version I'm currently using. 

Cheers,
Markus

### DIFF
--- a/Sources/Container.swift
+++ b/Sources/Container.swift
@@ -31,6 +31,7 @@ public final class Container {
     private let parent: Container?
     private var resolutionPool = ResolutionPool()
     private var properties = [String:AnyObject]()
+    private var pendingCompletes: [()->()] = []
     internal let lock: SpinLock // Used by SynchronizedResolver.
     
     /// Instantiates a `Container` with its parent `Container`. The parent is optional.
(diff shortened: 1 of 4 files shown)


yoichitgy:
Thanks @marcorei for the great idea to fix the issue. I was thinking it was difficult fix it, but you got it. Let me check the changes, and I'll come back soon.

yoichitgy:
Thank you @marcorei for fixing the difficult problem😃💡👍 I'll release an update soon.

Issue #125, opened 2016-07-11
Dependencies of ViewControllers are getting injected multiple times using reference Storyboards


phillippbertram:
#### Info

```
Swinject Version: 1.1.2
Xcode Version: Version 7.3.1 (7D1014)
```

#### Steps to reproduce

1. Instantiate Initial RootViewController described in [Without SwinjectStoryboard](https://github.com/Swinject/Swinject#without-swinjectstoryboard)
2. embed the template `ViewController` in a `UITabbarController`
3. create another storyboard called `Second.storyboard`
4. create a new view controller called `SecondViewController`
5. make a connection between the `UITabbarController` in `Main.storyboard` and the `SecondViewController` in `Second.storyboard`
6. register ViewControllers and dependencies like the following code example
7. Run the Application

#### Example Configuration

```swift
@UIApplicationMain
  class AppDelegate: UIResponder, UIApplicationDelegate {

    var window: UIWindow?

    let container = SwinjectStoryboard.defaultContainer
    
    func application(application: UIApplication, didFinishLaunchingWithOptions launchOptions: [NSObject: AnyObject]?) -> Bool {
        // Override point for customization after application launch.
            
        
        // Configure Swinject
        
        container.registerForStoryboard(CustomTabbarController.self) {
            r, vc in
            print("registerForStoryboard: CustomTabbarController")
            vc.viewModel = r.resolve(TabbarViewModel.self)
        }
        
        container.register(TabbarViewModel.self) {
            r in
            print("register: TabbarViewModel")
            return TabbarViewModel()
        }
        
        container.registerForStoryboard(FirstViewController.self) {
            r, vc in
            print("registerForStoryboard: FirstViewController")
            vc.viewModel = r.resolve(FirstViewModel.self)
        }
        
        container.register(FirstViewModel.self) {
            r in
            print("register: FirstViewModel")
            return FirstViewModel()
        }
        
        container.registerForStoryboard(SecondViewController.self) {
            r, vc in
            print("registerForStoryboard: SecondViewController")
            vc.viewModel = r.resolve(SecondViewModel.self)
        }
        
        container.register(SecondViewModel.self) {
            r in
            print("register: SecondViewModel")
            return SecondViewModel()
        }
        
        setupInitialView()
        
        return true
    }

    private func setupInitialView() {
        
        // We need to setup our view und rootViewController manually, so we
        // can use the SwinjectStoryboard for our dependency resolving mechanism
        
[...]

Issue #15, opened 2015-10-14
Enable "Require Only App-Extension-Safe API" to remove warning linking against dylib in application extensions


admeister:
Can the "Require Only App-Extension-Safe API" flag be enabled to stop the warning "linking against dylib not safe for use in application extensions" when referencing Swinject in an extension.

yoichitgy:
I checked v0.3 and I think the flog is set to YES. My misunderstanding??

yoichitgy:
Reference document:
https://developer.apple.com/library/watchos/documentation/General/Conceptual/ExtensibilityPG/ExtensionScenarios.html

admeister:
My bad, I was still on 0.2.3.

Great framework btw ;)

yoichitgy:
Ok, thanks.

I started using [jspahrsummers/xcconfigs](https://github.com/jspahrsummers/xcconfigs), and the flag was set to YES unconsciously and greatly.

PR #144, opened 2016-09-10
Storyboard reference bugfix

### LINKED ISSUE #128: Dependencies of non-root UIViewControllers are not getting storyboard injected
Xcode version: 7.3
Swinject version: 1.1.1

I've got a series of view controllers (drill down tableview) laid out in a Storyboard. The initial view controller in the storyboard is a navigation controller, with `ViewControllerA` as the root view controller. `ViewControllerA` is a static table view controller, and when one of its cells is selected, a segue is launched for `ViewControllerB` (drilling down). 

Both `ViewControllerA` and `ViewControllerB` are being injected with dependencies, but I'm only seeing `ViewControllerA`'s factory getting called. I've also double checked that both view controllers have the `Custom Class` field set properly in IB.

This storyboard is referenced from my main storyboard (The main storyboard is a tab bar controller). Let me know if you require any more information.

Assembly:

    class UIAssembly: AssemblyType {
        
            container.registerForStoryboard(ViewControllerA.self) { r, c in
                c.dependency = r.resolve(Dependency.self)! // Gets run
            }
            container.registerForStoryboard(ViewControllerB.self) { r, c in
                c.depencency = r.resolve(Dependency.self)! // Does not get run
            }
        }
    }

screen shot 2016-07-20 at 2 01 43 pm

jakubvano:
Ports [SwinjectStoryboard #15](https://github.com/Swinject/SwinjectStoryboard/pull/15) to v1.

Fixes #128, #125

### DIFF
--- a/Sources/SwinjectStoryboard/Storyboard+Swizzling.swift
+++ b/Sources/SwinjectStoryboard/Storyboard+Swizzling.swift
@@ -44,7 +44,11 @@ extension Storyboard {
     private class func swinject_storyboardWithName(name: String, bundle storyboardBundleOrNil: NSBundle) -> Storyboard {
         if self === Storyboard.self {
             // Instantiate SwinjectStoryboard if UI/NSStoryboard is tried to be instantiated.
-            return SwinjectStoryboard.create(name: name, bundle: storyboardBundleOrNil)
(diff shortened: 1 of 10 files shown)


yoichitgy:
Thank you for your great work to fix the issues👍👍👍

PR #145, opened 2016-09-10
Update for Xcode 8 GM with Swift 2.3

yoichitgy:
Updated the project for Xcode 8 GM with Swift 2.3:

- Updated depending test libraries (Quick and Nimble) to support Xcode 8 with Swift 2.3.
- Updated the project settings as Xcode 8 GM suggested.
- Updated Travis settings to use Xcode 8 image.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,9 @@ env:
     - TVOS_SDK=appletvsimulator10.0
     - WATCHOS_SDK=watchsimulator3.0
   matrix:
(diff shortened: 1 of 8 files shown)


yoichitgy:
In the Travis settings, a matrix using an iOS 8 device was commented out because Xcode 8 GM looked having a problem to start unit tests on an iOS 8 simulator. Only when an iOS 8 device is selected for test, not only in Travis but also local environment, unit tests do not start running. It looks a problem of Xcode.

The test using iOS 8 device should be reverted later when Apple fixes the problem.

PR #149, opened 2016-09-11
Update README to instruct which version of Swinject to use

yoichitgy:
@jakubvano, I updated README to tell which version of Swinject (v1 or v2) to use. Do you think my update is clear enough?

If it's ok, merge this PR please? If not, your advice is appreciated😃


jakubvano:
👍 I think it is understandable. Especially liked the `#Uncoment ..` - it is nice and succinct way of highlighting SwinjectStoryboard usage.

yoichitgy:
Thanks😃

PR #159, opened 2016-09-26
Fix dependency cycle injection

### LINKED ISSUE #158: Problem with postponed `initCompleted` calls
In the [1.1.4](https://github.com/Swinject/Swinject/releases/tag/1.1.4), calls of `initCompleted` are postponed (see implementation https://github.com/Swinject/Swinject/pull/134/commits/35011957dad63cf11c0e06c4b7ab3d4a91bc9ef2#diff-e549a00f41e3fe6ad1e35a13ebcaddb0L202).

Let's assume that we have `A` which has dependency `B`, `B` which has dependency `C` and `C` which has dependency `B`. 

`A`<-`B`<->`C`

```Swift
container.register(A.self) { r in
    return A(b: r.resolve(B.self)!)
}

container.register(B.self) { r in
    return B()
}.initCompleted { (r, b) in
    b.c = r.resolve(C.self)!
}

container.register(C.self) { r in
    return C(b: r.resolve(B.self)!)
}
```

If we want to call method `foo` of  `B` inside `A`'s initializer, we'll get an error, because dependency `C` of `B` is not initialized yet.

```Swift
class A {
    init(b: B) {
        b.foo()
    }
}

class B {
    var c: C! {
        didSet {
            print("C has been set")
        }
    }
    
    func foo() {
        print(c)
    }
}

class C {
    let b: B
    init(b: B) {
        self.b = b
    }
}
```

Above code was working in [1.1.2](https://github.com/Swinject/Swinject/releases/tag/1.1.2), because `initCompleted` was called at the end of resolving an instance.

Is this behavior expected in [1.1.4](https://github.com/Swinject/Swinject/releases/tag/1.1.4) or we can consider it as a bug?

marcorei:
Relevant #134 
Fixes #158 

Hi @yoichitgy 

Please take a look at a new approach to resolve dependency cycles without initiating objects twice.

The concept is only partially based on what I [wrote in the bug thread](https://github.com/Swinject/Swinject/issues/158#issuecomment-249362338). I could not think of a way to replace the Swift Optional with a reference type without breaking the API. So I took a different approach.

The PR is based on the premise that `initComplete` closures are not allowed to unwrap the Optionals resolved by the `Resolver`.  This should be in line with the [documentation](https://github.com/Swinject/Swinject/blob/v1/Documentation/CircularDependencies.md) but it will break code that tries to unwrap them (I had to remove force unwrapping from some Tests for this).

This PR enables Swinject to do two things:
- To investigate the dependencies needed by an `initComplete` closure using a [`DependencyAnalyzer`](https://github.com/marcorei/Swinject/blob/fix-dependency-cycle-injection/Sources/DependencyAnalyzer.swift) as the `Resolver`. It works by returning `nil` and saving the requested types.
- To call `initComplete` closures that only create new objects which are not in the process of being created.
- Then after each Objects initialization, pending `initComplete`s are checked and executed once all dependencies are created (that is: not currently being created).

The code might be rough by I though to pitch this idea to you before investing more time in this concept. 

Tests: 
- I added the case @petalvlad was describing.
- A case testing if storyboards are injected only once now fails because of the additional analysis run – I could not yet look into that issue.
- I had to remove force unwrapping from `initComplete` closures in tests.

Looking forward to your feedback!

Cheers,
Markus

### DIFF
--- a/Sources/Container.swift
+++ b/Sources/Container.swift
@@ -32,6 +32,8 @@ public final class Container {
     private var resolutionPool = ResolutionPool()
     private var properties = [String:AnyObject]()
     internal let lock: SpinLock // Used by SynchronizedResolver.
+    private var invokerSet = Set<ServiceType>()
+    private var pendingCompleteClosures = [PendingCompleteClosure]()
     
     /// Instantiates a `Container` with its parent `Container`. The parent is optional.
     ///
@@ -160,12 +162,12 @@ extension Container: Resolvable {
                 }
                 
                 if ownEntry.instance == nil {
-                    ownEntry.instance = resolveEntry(entry, key: key, invoker: invoker) as Any
+                    ownEntry.instance = resolveEntry(entry, key: key, invoker: invoker) as Any?
                 }
                 resolvedInstance = ownEntry.instance as? Service
             case .Hierarchy:
                 if entry.instance == nil {
-                    entry.instance = resolveEntry(entry, key: key, invoker: invoker) as Any
+                    entry.instance = resolveEntry(entry, key: key, invoker: invoker) as Any?
                 }
                 resolvedInstance = entry.instance as? Service
             }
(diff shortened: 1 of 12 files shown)


yoichitgy:
@marcorei, thank you for your investigation and pull request to fix the issue👍👍👍 Adding the test case for #158 is also great!

Please give me time to understand the problem and resolution correctly. Your detailed comment is very helpful! I'll also try adding some test cases.

I'll try taking time by the weekend.

yoichitgy:
@marcorei I've read your code. It's a nice solution for the problem👍

One concern is side effect caused by the call of `initComplete` by `DependencyAnalyzer`. A unit test in SwinjectStoryboardSpec is failing because of the side effect. I think the side effect is related to the assumption that `initComplete` must not force-unwrap the resolved instance.

So, I think it's good to run the analyzer and pending completions code only if an option for it is specified. For example, `.Lazy` object scope or `circularDependencyAnalized` as below. (My preference is the latter one though I still didn't investigate much.) What do you think about the concern?

```swift
container.register(AType.self) { _ in AImpl() }
    .inObjectScope(.Lazy)
    .initCompleted { (r, a) in
        a.b = r.resolve(BType.self)
    }
```

or

```swift
container.register(AType.self) { _ in AImpl() }
    .initCompleted { (r, a) in
        a.b = r.resolve(BType.self)
    }.circularDependencyAnalized()
```

Here are the other tiny nitpicks:

- With one of the options above, it's nice to put the analyzer and pending completion code to a function to simplify `resolveEntry`.
- DependencyAnalyzer should be `internal` to hide our implementation.

@jakubvano, what do you think about the problem and resolution?

Though we have the concern, we are getting closer to a good resolution. Thank you @marcorei😃

marcorei:
Running the analyzer optionally is a great idea in my opinion. I much prefer `circularDependencyAnalyzed()` because it can be used together with `.Container` or `.Hierarchy`.

jakubvano:
@yoichitgy My understanding of the problem is, that there are two premisies we might want to guarantee:

1. **No throw-away objects will be created during graph resolution** (i.e. temporary objects which are destroyed after resolution)
 - this was problem described in #134 
 - this is relevant when there are side-effects of creating given object (i.e time consuming operation, or some interaction with passed dependencies)

2. **Injected dependencies will be fully initialised**, except ones which are part of the same dep. cycle as given object
 - AFAIK this is currently not possible to achieve when object is part of a dependency cycle
 - this covers #158

Initial solution (with immediate `initComplete` calls) fulfilled 2 but not 1, solution in #134 fulfilled 1 but not 2.

Proposed solution fulfills 2 as well, but with couple of new issues:
 - due to multiple `initComplete` invocations 1 does not hold for object scope `.None`dependencies. This is IMO not relevant, but should be mentioned in docs if we choose to go with this
 - due to multiple `initComplete` invocations, property dependencies will be assigned multiple times.
 - forced unrwapping is not allowed in `initCompleted` closures

As you mentioned, if we do not want to break current API we should make this an optional feature - I would suggest having #134 as default implementation and adding something like `.attemptCompleteInitialisation()` for opting to use `DependencyAnalyser` feature.

However, I do think that proposed solution should be a default behaviour, as both 1 and 2, seem like natural expectations for DI framework. Maybe add opt-in feature in `1.0` version and make it default in `2.0`?

jakubvano:
Thinking about this some more, maybe initial solution with immediate `initCompleted` calls was fine after all: it is much simpler, and in most cases (at least in my experience) "throw-away" instances do not cause problems in properly designed architecture.

And if they do, you can work around them by injecting both parts of dependency cycle in `initCompleted` closures, i.e:
```
container.register(A.self) { _ in A() }
    .initCompleted { r, a in a.b = r.resolve(B.self)! }

container.register(B.self) { _ in B() }
    .initCompleted { r, b in b.a = r.resolve(A.self)! }
```
instead of
```
container.register(A.self) { r in A(b: r.resolve(B.self)!) }

container.register(B.self) { _ in B() }
    .initCompleted { r, b in b.a = r.resolve(A.self)! }
```

yoichitgy:
Thanks @jakubvano for your deep investigation and understanding of the problem🙇 You structured my ambiguous concern. Invoking `initCompleted` is important for both of Swinject code and user's code. Swinject should be simple, stable, robust and consistent because DI framework is a fundamental framework for an app.

I would like to conclude we revert #134, and add documentation about the workaround @jakubvano  wrote. @marcorei, would you agree?

The idea using DependencyAnalyzer is technically very interesting, but a little bit experimental to add to Swinject Core. It may be good to add that kind of idea to a Swinject extension hosted in another repository. @marcorei, I really appreciate your pull request, idea and reference to Dagger😃 If you are interested, I appreciate another idea or pull request to a different problem in Swinject Core or a new extension!

yoichitgy:
This is off topic, but I want to simplify my code implementing the feature of parent container and hierarchy scope. It is the most complicated part in Swinject...

marcorei:
@yoichitgy I agree, the solution is rather complex for a small benefit. Working on it was fun nevertheless!

yoichitgy:
Thanks @marcorei! Ok, we will revert #134 and following some adjustment commits I did, and document the workaround.

yoichitgy:
#134 has been reverted by #164.

Issue #154, opened 2016-09-22
Add a way to log failing resolution


yoichitgy:
Idea from: https://github.com/Swinject/Swinject/issues/153#issuecomment-248351479

> as user-friendliness usually beats documentation, maybe we could implement some run-time warning messages similar to [checkedResolved](https://github.com/Swinject/SwinjectAutoregistration/blob/master/Sources/AutoRegistration.swift#L682), which would somehow log that resolve failed, what arguments were given, and maybe some suggestion to what might be the problem

```
Resolving of MDFDataSource failed.
Name: -
Arguments: UITableView!, MDFFetchedResultsController

Found similar registration for MDFDataSource:
Name: -
Arguments: UITableView, MDFFetchedResultsController
```

jakubvano:
Implemented by #160

PR #160, opened 2016-09-30
Log resolution failure

jakubvano:
Implements #154 

When resolution fails, this logs description of expected parameters, and lists all available registrations for given service type.

Example behaviour:
```
let container = Container()
container.register(Int.self, name: "A") { _ in 0 }
container.register(Int.self) { (_, a: String, b: Double) in 0 }
container.register(Double.self) { _ in 0 }

container.resolve(Int.self)
```
```
Swinject: Resolution failed. Expected registration:
	{ Service: Int, Factory: (Resolver) -> Int }
Available registrations:
	{ Service: Int, Name: "A", Factory: (Resolver) -> Int, ObjectScope: graph }
	{ Service: Int, Factory: ((Resolver, String, Double)) -> Int, ObjectScope: graph }
```

Disabling / customising of this feature can be done by changing `Container.loggingFunction` variable.


yoichitgy:
It's a very cool feature to have. Thank you for implementing it👍 As you told, it's helpful to notice the type mismatch problems!

Issue #106, opened 2016-06-08
Allow for multiple parents


amccarri:
I have a layered app architecture that divides responsibilities across several frameworks.  At the root level is an framework that depends on underlying peer frameworks.  Each one of those has it's own DI container.  I would like to expose services defined in all the underlying containers to the overall app.  This is currently not possible without creating an unnecessary dependency between peer frameworks since the container hierarchy only allows a single parent.

It would be nice if I could register multiple parent containers.

yoichitgy:
Thanks @amccarri for the nice idea. Regarding parent/child relationship of containers, I would like to implement #80 first. It looks we need a good idea for implementation not to conflict with #80.

jakubvano:
@amccarri I have been going over issues relating to parent-child container relationships - What you are describing looks like perfect usecase for `AssemblyType`s and `Assembler` - instead of exposing `Container` from individual frameworks you could expose an `AssemblyType` for each and then combine them into a single `Assembler`.

Or am I missing something there?

yoichitgy:
I've been thinking about this, and Assembler looks working in this case. Adding multiple parents has a kind of [diamond problem](https://en.wikipedia.org/wiki/Multiple_inheritance) like multiple super class inheritance in C++ or similar.

yoichitgy:
Closing this issue as `wontfix`. Please feel free to reopen this issue if you have a strong reason and good implementation to add this feature.

Issue #98, opened 2016-05-16
Usecase for multiple containers


jakubvano:
Original post in #80 

> For a while I was viewing parent - child relationship as strange too, but I grew to like it and now find it useful for managing scopes with different lifetime within application.
> 
> On the other hand, I am probably missing usefulness of having multiple Containers / Assemblers when you only want to keep single reference to the parent - AFAIK at that point it doesn't matter whether services are registered on parent or child, as you can only call resolve() on parent. So why not register them all on parent Container? 
> I'd love if you could clarify this for me.

jakubvano:
To clarify what I mean by scope management:

 I do use multiple containers with parent-child relationship on them when I need some "singletons" (`.inObjectScope(.Hierarchy)`) to be destroyed during application life. (E.g. objects holding user-sensitive data after logout). I that case, I have an `applicationContainer` as a parent of `sessionContainer`. After logout `sessionContainer` is destroyed and re-created on login.

However, I am wondering If there are other usecases for multiple `Container`s, which would maybe benefit from different nature of parent - child relationship than is currently implemented

yoichitgy:
Thanks for creating the issue here. I'll answer shortly😃

yoichitgy:
Container hierarchy is used for lifetime management of singletons as you said. Another example is iTunes app for Mac. It can have something like `applicationContainer`, `MusicContainer`, `IPhoneAppsContainer`, etc to manage memory usage. Another use case is just for grouping, but it might be a preference.

Issue #80, opened 2016-04-08
Enhancement of API for container parent-child ownership


thalmicMark:
I'm kind of perplexed on how to manage a single `Container`/`Assembler` with several children `Container`s/`Assembler`s. This is how it currently works:

```
1. The parent Container/Assembler is created.
2. The children Containers/Assemblers are created, passing the parent container in their constructors.
3. The children retain a strong reference to their parents.
4. Something needs to hold on to all Containers/Assemblers (parents and children).
```

Rather than have the children keep a strong reference to its parent, why not have the parent keep a strong reference to its child, and the child keeps a weak reference to its parent? What this would allow for:

```
1. The parent Container/Assembler is created.
2. The children Containers/Assemblers are created.
3. The children are added to the parent by calling parent.addChild(child).
4. A single reference to the parent container is needed in order to keep everything alive.
```

The `Container`/`Assembler` constructor that takes a parent parameter can still exist (preserving backwards compatibility) by calling `parent.addChild(child)` instead of setting the property directly.

yoichitgy:
Thank you for posting the issue😃 I understand what you mentioned. I made the way to add parent/child relationship just as early version of [Funq](https://funq.codeplex.com) did without thinking so much. (Current version of Funq is different.)

I also thought the relationship was strange. I think we don't have to care about backward compatibility for the meaningful change.

jakubvano:
For a while I was viewing parent - child relationship as strange too, but I grew to like it and now find it useful for managing scopes with different lifetime within application. 

On the other hand, I am probably missing usefulness of having multiple `Container`s / `Assembler`s when you only want to keep single reference to the parent - AFAIK at that point it doesn't matter whether services are registered on parent or child, as you can only call `resolve()` on parent. So why not register them all on parent `Container`? 
I'd love if you could clarify this for me.

yoichitgy:
@jakubvano would you open a new issue for your question with a good title? Just copying your comment to it, and deleting your comment from this issue are ok.

jakubvano:
Created #98 , also elaborated a bit on how I currently use multiple containers.

yoichitgy:
`CreateChildContainer` of Funq Container is a good example to change the API.
https://funq.codeplex.com/SourceControl/latest#src/Core/Funq/Container.cs

jakubvano:
Closing due to inactivity

gfpoliva:
@jakubvano was this implemented?

jakubvano:
[...]

PR #165, opened 2016-10-07
Customizable object scopes

jakubvano:
Relates to #80, #98, #106.

I've been thinking about parent-child container relationship, and realised that one of the difficulties of implementing e.g. `.hierarchy` scope, or some more complex object life time management is the fact that `Container` manages both service registrations and service instances.

So I thought to move responsibility of managing instance persistence to `ObjectScope`. It is currently only a draft, and would ofc require some more work. @yoichitgy What do you think? Could this be a way to go or is it too much of a breaking change?

Some implications of this paradigm:
- `ObjectScope.hierarchy` doesn't make sense, and `.singleton` is more fitting than `.container`
- parent container might be redundant, as scopes would (probably) be shared among all containers
- This enables custom scopes, e.g. if I had `SessionController` holding user-sensitive data which should be destroyed after logout, I could implement it like:

```
extension ObjectScope {
    static let user = ObjectScope(parent: .singleton, name: "user")
}

container.register(SessionController.self) { _ in SessionController() }
    .inObjectScope(.user)

func logout() {
    ...
    ObjectScope.user.discard()
}
```


yoichitgy:
I'll reply shortly.

yoichitgy:
Thanks for the great idea💡 Swinject v2 is still beta and it's a good opportunity to add breaking changes for great enhancements like you are thinking. The parent container and hierarchy scope made the implementation complicated, causing difficulty to add the other nice features.

It's nice to move the responsibility to ObjectScope because, as you said, it's extendable. It's also nice if experimental scopes can be added to another repository in Swinject organization, or third party repositories.

The `.user` scope is nice. I also wanted to add `.thread` or `.session` scopes for server-side use cases.

I chose the wording `.container` because Funq used the wording and Singleton pattern like `UIApplication.shared` is an anti pattern in DI, though the lifecycle of objects in `.singleton` scope is tied to their container. I wanted to avoid confusion for Swinject beginner users for the wording, though some other DI frameworks (including the other languages) name it singleton scope. By the way, `.none` is also a candidate to rename to clarify.

So, let's migrate to the better architecture you are writing in v2 beta. We can provide better scopes than `.hierarchy`.

haitran2011:
ok

jakubvano:
After several iterations I've finally come to an implementation of custom scopes I'm more-or-less satisfied with.

It provides `graph`, `none`, and `container` scopes, with easy addition of custom scopes, for example:

`user` scope:
```swift
extension ObjectScope {
    static let user = ObjectScope(storageFactory: PermanentStorage.init)
}
...
func logout() { container.resetObjectScope(.user) }
```

`weak` scope (#185):
```swift
extension ObjectScope {
    static let weak = ObjectScope(storageFactory: WeakStorage.init)
}
final class WeakStorage: InstanceStorage {
    class Wrapper {
        let value: Any
        init(_ value: Any) { self.value = value }
    }

    private weak var wrapper: Wrapper?
    var instance: Any? {
        get { return wrapper?.value }
        set { wrapper = newValue.map(Wrapper.init) }
    }
}
```
or `thread` scope:
```swift
extension ObjectScope {
    static let thread = ObjectScope(storageFactory: ThreadUniqueStorage.init)
}
final class ThreadUniqueStorage: InstanceStorage {
    var instances = [ObjectIdentifier: Any]()
    var instance: Any? {
        get { return instances[ObjectIdentifier(Thread.current)] }
        set { instances[ObjectIdentifier(Thread.current)] = newValue }
    }
}
```

Possible issue:
 -  I chose the`container` name as it has semantics of "instaces are persited for the entire lifetime of a container they are registered in", however this is in effect the same as old `.hierarchy` scope, so maybe it should be renamed.
- there is no equivalent of  old `container` scope, but IMO is not needed as container-based lifetime management of instances can be more easily expressed with custom scopes.
- I haven't changed the nature of parent-child relationship as I couldn't identify exact problems with current relationship, and removal of old `container` context IMO sufficiently reduces complexity.

yoichitgy:
This feature removed the `switch` statement for scopes from `Container`, and its implementation was simplified. Using `ObjectScope` to create a `storage` lazily for a service entry is extensible and nice idea👍 Overall, I think it's great to migrate to the new scope model as Swinject v2.

- I think the concept of the old `.hierarchy` and `.container` scopes was complicated. When a parent-child relationship is created, taking the old `.hierarchy` scope (the new `.container` scope) is simple and makes sense.
- I think the scope name `.container` is ok because it is specified when a service is registered to a container. If the container has a child container, the child gets the instance stored in the parent container where `.container` scope is specified. It makes sense for me.
- If the old `.container` scope is really needed, it can be achieved by adding the same service registration to both parent and child containers. So, removing the old `.container` can be no problem. The advantage to get things much simpler is more important.
- `ObjectScope.none` looks good to be renamed to `.transient` as you named `TransientStorage` because `.none` doesn't make sense (though I named it in v1).

I have a nitpicking question.

- `resetObjectScope` of a container calls the method of its parent. Should it call the parent's? (I don't mean it shouldn't. I'm thinking which is more reasonable whether calling it or not.) What is the reason why the child call of `resetObjectScope` calls the parent one?

jakubvano:
>ObjectScope.none looks good to be renamed to .transient as you named TransientStorage because .none doesn't make sense (though I named it in v1).

I agree, `.transient` is more descriptive.

>resetObjectScope of a container calls the method of its parent. Should it call the parent's? (I don't mean it shouldn't. I'm thinking which is more reasonable whether calling it or not.) What is the reason why the child call of resetObjectScope calls the parent one?

Main reason is `.graph` object scope, which needs to be reset after object graph is resolved. - if `resetObjectScope` did not reset parent, we would need different mechanism for `.graph` object scope, which would complicate implementation.
I can't come up with other arguments for either option, mostly because with custom scopes, I don't see reason for using more than one container - grouping can be achieved via `AssemblyType` and lifetime management via custom scopes. 
However, if you can see some scenarios in which not resetting the parent might be useful, I wouldn't mind changing the behaviour.

yoichitgy:
Thank you for explaining the reason about `resetObjectScope`. I understood it😃 Let's move on with this implementation. (So far, I don't see any problems. I'll only ask you for some discussion if I find a problem for resetting parent's before we release Swinject v2 official version.)

As far as I heard, some people prefer using `AssemblyType` and some people prefer managing container groups by themselves with/without parent relationship. I think dropping parent relationship might be a too large breaking change.

jakubvano:
I've renamed to `.transient` scope and updated documentation, so PR should be ready to merge.

I've also considered adding `.weak` object scope as discussed in #185, but it is probably better to do in separate PR.

yoichitgy:
Swinject became not only more extendable but also simpler to maintain👍

basalphenaar:
I couldn't help but check out the latest commit of Swinject to try this new feature, it can be really useful for the app I'm currently building (to manage user sessions). I'm also using SwinjectStoryboard and it doesn't seem to be working with that yet (I get a lot of segmentation fault errors when building my project). Did anybody get this running in combination with SwinjectStoryboard or is this not possible yet?

jakubvano:
@basalphenaar currently it wont work with `SwinjectStoryboard`, because it has specified dependency on older `Swinject` release, thus two versions of Swinject framework being used (one in app one as dependency of `SwinjectStoryboard`) are not compatible.

basalphenaar:
@jakubvano Thanks for your answer, indeed Carthage was also giving some conflict errors when using the latest commit of Swinject together with SwinjectStoryboard 1.0.0-beta2. No problem, I'll think of some other solution for now :)

PR #174, opened 2016-10-15
Update .travis.yml.

yoichitgy:
- Replace Carthage with `git submodule update` because using Carthage is redundant.
- Add xcpretty again because random test failure on travis didn't relate to xcpretty.


yoichitgy:
@jakubvano, I made this simple change as a pull request to let you know.

Issue #162, opened 2016-10-05
Xcode still asks for Swift conversion of Swinject


sialcasa:
I've created a new project from scratch, did a pod init and changed the podfile to:

```
source 'https://github.com/CocoaPods/Specs.git'
platform :ios, '8.4'
use_frameworks!

target "mobiped" do
  pod 'Swinject', '2.0.0-beta.2'
end
```

Then I did a pod install and opened the workspace in XCode. 

Now the problem: XCode still askes me to convert swinject to Swift 2.3 / 3, otherwise I get compile errors.

What did I wrong?

yoichitgy:
Please make sure you use CocoaPods 1.1.0.rc.2 or later.

yoichitgy:
@sialcasa, resolved the issue?

sialcasa:
Yes thank you

PR #175, opened 2016-10-15
Informing users of requirement of required init when using Swinject with UIViewControllers

Sajjon:
resolving a UIViewController that lacks required init fails, this texts informs about this.
As per my [StackOverflow reply](http://stackoverflow.com/a/38370822/1311272)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Dependency injection (DI) is a software design pattern that implements Inversion
   - Xcode 8.0+
 - Carthage 0.18+ (if you use)
 - CocoaPods 1.1.0.rc.2+ (if you use)
-- When using Swinject for resolving ViewControllers, your `UIViewController` subclass *must* implement the `required init?(coder aDecoder: NSCoder)` initializer.
+- When using Swinject for resolving ViewControllers, your `UIViewController` subclass **must** implement the `required init?(coder aDecoder: NSCoder)` initializer.
 
 ## Installation
 
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ Dependency injection (DI) is a software design pattern that implements Inversion
   - Xcode 8.0+
 - Carthage 0.18+ (if you use)
 - CocoaPods 1.1.0.rc.2+ (if you use)
+- When using Swinject for resolving ViewControllers, your `UIViewController` subclass *must* implement the `required init?(coder aDecoder: NSCoder)` initializer.
 
 ## Installation
 


jakubvano:
👍  Thanks for PR! I'd like to clarify though - are there some specific cases in which `init(coder:)` is needed? Because current tests pass even with commented-out initialiser on view controllers.

Sajjon:
@jakubvano I have a project where I am not using storyboard nor xib, and I have two viewcontrollers, `class VehicleVC: UIViewController` and `class CarVC: VehicleVC` where CarVC can make use of init in VehicleVC.

If both VCs are registered to the `container` and then initialized using `container.resolve` then it will fail for CarVC, if it lacks `required init(coder)`. 

Just adding the required init and the init you wanna use (override) and then calling super, fixes this.

jakubvano:
My bad, as stack overflow question was related to `SwinejctStoryboard` i assumed that is where issue was. However, I can't seem to reproduce this faulty behaviour. Could you post minimal example? Or can it be somehow reproduced in the project you have linked in SO answer?

Sajjon:
@jakubvano sure, which version/branch/commit should I be using for Swift 3? is it still `pod 'Swinject', '2.0.0-beta.2'` ? Because that still leads to a migration guide, with some changes in the swinject pod. I used `c598e7a` in the project I had problems with. Which is the most correct? :)

jakubvano:
`2.0.0-beta.2' should be correct - are u using `1.1.0.rc.2` (or later) version of cocoapods?

Sajjon:
Ah, 
```
› pod --version
1.0.1
```
So if I upgrade to 1.1.0.rc.2 it will work? :)

jakubvano:
It should - see #162

Sajjon:
Okay I will update PR and mention this in the README under Swift 3, because it is not mentioned :)

Sajjon:
@jakubvano Ah nice! Now it works! https://github.com/Sajjon/SwinjectFailsIfNoRequiredInit
So missleading git hub repo name :P since it is working! Must have been fixed lately.

So should we maybe mention that it is required to have pod version `1.1.0.rc.2`?

jakubvano:
That is good to know 😄 

About cocoapods version - it is not exclusive for Swinject (AFAIK it affects all Swift 3 pods), but mentioning it might be helpful for integration. And once final `1.1.0` version is released we can update / remove that reference. 

Feel free to open new pull request with that change (as it does not directly relate to this topic) and I'll merge it

yoichitgy:
README describes "CocoaPods 1.1.0.rc.2+ (if you use)" but "+" might be confusing. It looks good to change it to "or later" as well as the other requirements. Also, making **1.1.0.rc.2** bold looks also good if @Sajjon thinks it is helpful for new users.

What do you think, @Sajjon? Your kind pull request is welcome to improve the README for new users of Swinject😃

yoichitgy:
Thanks @Sajjon for opening a pull request to improve Swinject project. I'm going to close this pull request, but I always appreciate your pull requests to enhance Swinject code or documentation😃

PR #178, opened 2016-10-20
Added a bool parameter to Container init to enable/disable debug logging

tralf:
We use Swinject at our project and very appreciate your work. 

In the project we intentionally have some dependencies unregistered at an assembler. And after the last release it started bombard with error logs that some instances unregistered. As we're aware about it, we tried to find the way to disable the log, but there was no it.

So I added a minor change to disable the logger.

### DIFF
--- a/Sources/Container.swift
+++ b/Sources/Container.swift
@@ -27,10 +27,10 @@ public final class Container {
     fileprivate var services = [ServiceKey: ServiceEntryType]()
     fileprivate let parent: Container?
     fileprivate var resolutionPool = ResolutionPool()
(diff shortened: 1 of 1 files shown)


jakubvano:
Thanks for the PR! 👍  However there already is a way to disable logging:
```
Container.loggingFunction = nil
```
as documented [here](https://github.com/Swinject/Swinject/blob/v1/Documentation/Misc.md#resolution-failure-logging). By mistake documentation changes from` v1` were not merged to `master` branch - will be fixed by #179

tralf:
Awesome, thank you!

PR #182, opened 2016-10-28
Fix storyboard injection

### LINKED ISSUE #177: isCreatingStoryboardReference -> a question
Today I have been investigating the source of the problem I asked about in the following Stack Overflow link.

http://stackoverflow.com/questions/40050100/dependency-injection-inconsistency-in-differing-viewcontrollers-in-swinject-pos

The particular set-up I have here is a TabBarController, containing four UIPageViewControllers, each of which has it's own set of UIViewControllers. When these UI elements pass through SwinjectStoryboard.instantiateViewController this is the printed debug output in the order it happens:

```
[line 81] instantiateViewController id: TabBarController
[line 86] instantiateViewController id: TabBarController 
[line 88] instantiateViewController id: TabBarController
...
[line 81] instantiateViewController id: UIPageViewController-gL2-Zr-HmB
[line 81] instantiateViewController id: DashboardViewController
[line 86] instantiateViewController id: DashboardViewController 
[line 86] instantiateViewController id: UIPageViewController-gL2-Zr-HmB, vc: <JBM.AnalysisPageViewController>
[line 88] instantiateViewController id: UIPageViewController-gL2-Zr-HmB
```

_line 81 is the start of the **instantiateViewController** function, line 86 is just before **if !SwinjectStoryboard.isCreatingStoryboardReference**, and line 88 is inside **if !SwinjectStoryboard.isCreatingStoryboardReference**._

So you see that the UIPageViewController pauses halfway through its own instantiation to allow the its default page "DashboardViewController" (as set in Main.storyboard) to instantiate itself. And this means that **SwinjectStoryboard.isCreatingStoryboardReference** is true, and therefore that DashboardViewController is not injected, even though it has been correctly registered for injection thus:

```
extension SwinjectStoryboard
{
    class func setup ()
    {
    defaultContainer.registerForStoryboard( DashboardViewController.self )
    {
        r, vc in... etc.
    }
```

What should be done about this, do we think? I myself am not familiar enough with the internals of Swinject to know what the correct way forward is.

:/

jakubvano:
Resolves #177

### DIFF
--- a/Sources/RegistrationNameAssociatable.swift
+++ b/Sources/RegistrationNameAssociatable.swift
@@ -6,19 +6,6 @@
 //  Copyright © 2015 Swinject Contributors. All rights reserved.
 //
 
-import Foundation
-import ObjectiveC
-
 internal protocol RegistrationNameAssociatable: AnyObject {
     var swinjectRegistrationName: String? { get set }
-}
-
-extension RegistrationNameAssociatable {
-    internal func getAssociatedString(key key: UnsafePointer<Void>) -> String? {
-        return objc_getAssociatedObject(self, key) as? String
-    }
-    
-    internal func setAssociatedString(string: String?, key: UnsafePointer<Void>) {
-        objc_setAssociatedObject(self, key, string, objc_AssociationPolicy.OBJC_ASSOCIATION_COPY)
-    }
 }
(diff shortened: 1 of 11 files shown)


yoichitgy:
Restarted failed CI.

yoichitgy:
Thanks @jakubvano for fixing the issue👍

PR #196, opened 2016-11-25
Swift Package Manager and Linux support

yoichitgy:
Swift Package Manager updated its source layout for unit tests. This change caused build errors on `swift build` command to build Swinject. To fix the issue, I moved unit test files to follow the source layout and added an empty LinuxMain file.

Because of dependency of our tests on Quick and Nimble, still the tests cannot run on Linux (nor Mac with `swift test` command), but at least `swift build` command works on Linux and Mac.


ikesyo:
Quick and Nimble surely work on Linux. See ReactiveSwift and https://github.com/ReactiveCocoa/ReactiveSwift/pull/86 as an example.

yoichitgy:
Thanks @ikesyo for the info about linux compatibility with Quick and Nimble. I had misunderstanding that only Quick supports (and Nimble does not support) Linux from the following Linux example.
https://github.com/Quick/QuickOnLinuxExample

yoichitgy:
`mv .Package.test.swift Package.swift` in [.travis.yml](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/.travis.yml) is a nice practice.

yoichitgy:
Added Linux support and SPM with unit tests using Quick and Nimble successfully. Thanks @ikesyo for your kind message to notify Quick and Nimble support Linux😃

I updated test matrix of Travis to run unit tests on Linux and various simulators on Mac using Xcode. Some open source projects run unit tests on Mac using Swift Package Manager, but I didn't add the case because I think adding various types and OS versions of simulators are more important, and tests on Mac with SPM is supposed to be ok if tests on Linux with SPM and on Mac with Xcode pass.

@jakubvano, would you review this pull request, and merge it if it looks ok?

.swiftlint.yml
```diff
@@ -6,6 +6,7 @@ disabled_rules:
   - opening_brace
   - valid_docs
   - type_name
+  - vertical_whitespace
```
yoichitgy: In my environment, `vertical_whitespace` rule generates warnings in Assembler.swift and Container.swift with Swiftlint v0.13.0. I thought it was ok to have 2 or more consecutive line breaks, so I excluded the rule rather than fixing the source code. I actually have no preference about this. If you like the rule, I'll remove excluding the rule and fix the source code.

Basically I agree that tests are also our codebase, but I excluded the tests from linting because we use Quick and our unit tests (or specs) have long `spec` functions. This is not my strong decision. It's ok to add a new issue to quit excluding the tests and fix some parts of .swiftlint.yml or test code.

jakubvano: I didn't have the latest `swiftlint` so violations did not show in my environment. Looking at those files, I don't think that multiple whitespaces help readability, so I would rather keep default settings for rule and remove whitespaces.

As for tests - it is IMO better to disable `function_length` rule then disable linting altogether (even Quick templates have it disabled by default 😃). I've created issue #197

README.md
```diff
@@ -8,9 +8,10 @@ Swinject
 [![Travis CI](https://travis-ci.org/Swinject/Swinject.svg?branch=master)](https://travis-ci.org/Swinject/Swinject)
 [![Carthage compatible](https://img.shields.io/badge/Carthage-compatible-4BC51D.svg?style=flat)](https://github.com/Carthage/Carthage)
 [![CocoaPods Version](https://img.shields.io/cocoapods/v/Swinject.svg?style=flat)](http://cocoapods.org/pods/Swinject)
+[![Swift PM compatible](https://img.shields.io/badge/SwiftPM-compatible-4BC51D.svg?style=flat)](https://github.com/Carthage/Carthage)
```
yoichitgy: Thank you for your careful eyes. For another reason, I dropped the Swift PM badge at its following commit (25934ab306680214213380b258109f2f4a0ef24c) to fit all badges in a line.

yoichitgy:
- I Removed exclusion of `vertical_whitespace` rule from Swiftlint settings as you mentioned in the review.
- `Swift PM` badge was already removed as mentioned in [my comment](https://github.com/Swinject/Swinject/pull/196#discussion_r90220532).

PR #198, opened 2016-11-30
Weak object scope

### LINKED ISSUE #185: Enhancement of API for weak singletons
I'd like to make a feature request for weak singletons. There are many cases in your app where you need to use a singleton, but the issue is the singleton will live on for the lifetime of the app.

Instead, I propose weak singletons. Essentially Swinject would maintain a weak reference to the class when it is resolved instead of a strong one. Anyone else who resolves the class while the first resolved instance exists will thus get that same instance. When all resolved instances are released, the singleton can then be de-initialized, since Swinject itself is not holding onto it. If someone else then resolves the class again, a new instance will be created.

Thoughts?

jakubvano:
Resolves #185.

### DIFF
--- a/Documentation/ObjectScopes.md
+++ b/Documentation/ObjectScopes.md
@@ -28,6 +28,12 @@ In `ObjectScope.container`, an instance provided by a container is shared within
 
 This scope is also known as _Singleton_ in other DI frameworks.
 
+### Weak
+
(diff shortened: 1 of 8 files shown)


yoichitgy:
It looks great👍👍👍 I merged it and fixed some conflicts with PR #196.

PR #202, opened 2016-12-02
Fix Any to AnyObject cast for linux environment

jakubvano:
The issue: On non-darwin platofrms `Any` is not castable to `AnyObject`, while on darwin platform conditional cast of `Any` to `AnyObject` produces warning.

### DIFF
--- a/Sources/InstanceStorage.swift
+++ b/Sources/InstanceStorage.swift
@@ -33,10 +33,17 @@ public final class TransientStorage: InstanceStorage {
 public final class WeakStorage: InstanceStorage {
     private weak var object: AnyObject?
 
+    #if os(Linux)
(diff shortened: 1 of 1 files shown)


yoichitgy:
Thank you for catching the problem for Linux👍 I also saw `newValue as AnyObject?` generates the error on Linux, and `newValue.flatMap { $0 as? AnyObject }` generates the warning on Mac. It is inconsistency of Swift lang implementation between Linux and Mac.

jakubvano:
AFAIK `Any` is castable to `AnyObject` only because of objC interoperability, so it makes sense that it does not work on non-darwin systems. From what I gather it is recommended not to use such features in multi-platform code, but here it cannot be avoided.

PR #214, opened 2017-01-01
added a graph to help you understand the DI Container

gonini:
These graph images helped me a lot when I learned about Android's dependency injection library, dagger.
Swinject's technical documentation is very nice.
I hope that the graph I created will be helpful in the technical documentation.
If it is not intuitive, we will recreate it.
Thank you.


gonini:
# DI Container

[Dependency injection](https://en.wikipedia.org/wiki/Dependency_injection) (DI) is a software design pattern that uses [inversion of control](https://en.wikipedia.org/wiki/Inversion_of_control) (IoC) for resolving dependencies. A DI _container_ manages the type dependencies of your system. First, you register the  types that should be resolved, with their dependencies. Then you use the DI container to get instances of those types whose dependencies are then automatically resolved by the DI container. In Swinject, the `Container` class represents the DI container.

Unfortunately, implementations of dependency injection often use slightly different terminology, which if not explained clearly in advance can cause confusion. In Swinject we define the following terms:

<p align="center">
  
</p>

* **Service**: A protocol defining an interface for a dependent type.
* **Component**: An actual type implementing a service.
* **Factory**: A function or closure instantiating a component.
* **Container**: A collection of component instances.

gonini:
The following changes have been made.
https://github.com/gonini/Swinject/blob/master/Documentation/DIContainer.md

jakubvano:
@gonini Thanks for the PR! 👍 
Could you link dagger graphs you were inspired by? It might help me to understand what kind of graphs people coming from other frameworks might be accustomed to.
That being said, I would consider including *Factory* instead of *Service* in the graph - IMO factories are quite important to understanding of how Swinject works. 

On related note - It probably wouldn't hurt to go through docs and make sure that terminology is consistent - I think that on multiple places *Registration* is beeing used instead of *Component*. @yoichitgy What do you think about those two?

gonini:
delighted to be of assistance. This is an [article](http://fernandocejas.com/2015/04/11/tasting-dagger-2-on-android/) about dagger2 that I've read. And the graph is an image. 
<p align="center">
  
</p>

yoichitgy:
Thanks @gonini for the pull request to improve the documentation, and the link to the great post to understand DI with Dagger👍

As @jakubvano wrote, it is a nice idea to add **Factory** to the image. Also consistency of the terminology is also important. I used **Component** meaning "implementation of interface", but the wording might be confusing to people who came from Dagger because it is used to mean **Injector**, which is a container in Swinject. So, changing **Component** to another word looks good IMO.

Anyway, I wrote the documentation in only limited time when I started Swinject project, though @jlyonsmith improved it a lot. It is nice to revise it especially about definition of or introduction to Dependency Injection.

@gonini, adding a picture to the documentation is a great idea to illustrate the concept of Dependency Injection. I think it is nice if the picture matches the context of the documentation. I have questions to @gonini. Did you think the following part of [the current documentation](https://github.com/Swinject/Swinject/blob/master/Documentation/DIContainer.md) was confusing? If so, which part was confusing and how?

> - Service: A protocol defining an interface for a dependent type.
> - Component: An actual type implementing a service.
> - Factory: A function or closure instantiating a component.
> - Container: A collection of component instances.

yoichitgy:
Hi @gonini. Thank you for your pull request to improve Swinject project👍 I'm going to close this issue for now, but please feel free to reopen this pull request if you would like to restart discussion to fit your pull request in the project!

PR #237, opened 2017-03-28
Fix SwiftLint errors with the latest version

dagio:
### :pushpin: References
* **Issue:** None
* **Related pull-requests:** None

### :tophat: What is the goal?

The latest version of SwiftLint (0.17.0) introduce a breaking change with rule `variable_name`.

### How is it being implemented?

Ignore the rule so the project can compile.

### DIFF
--- a/Sources/Container.swift
+++ b/Sources/Container.swift
@@ -119,6 +119,7 @@ public final class Container {
     ///
     /// - Returns: A registered `ServiceEntry` to configure more settings with method chaining.
     @discardableResult
+    // swiftlint:disable:next identifier_name
     public func _register<Service, Factory>(
         _ serviceType: Service.Type,
         factory: Factory,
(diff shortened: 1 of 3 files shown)


yoichitgy:
Thanks @dagio for catching the issue👍👍👍

IMO, it's good to update `.swiftlint.yml` to disable the lint to keep our source code cleaner. Would you disable `variable_name` in `.swiftlint.yml`❓

dagio:
It depends what strategy Swinject wants to adopt regarding new rules in general. This PR just make sure at least it can build.

jakubvano:
I would prefer the approach of disabling the rule for specific identifiers - disabling it for the entire project would imo be like saying "we do not care about identifier names at all".

yoichitgy:
I'm sorry that I took 10 days to come back. Let me just check whether SwiftLint has a way to disable a rule for specific identifiers. If it doesn't have it (and probably), I'll merge this pull request👍

yoichitgy:
Thanks @dagio for your pull request and @jakubvano for your idea👍 I checked some projects using SwiftLint, and disabling some exceptional cases works very well. Sorry, @dagio, for taking time to merge your pull request😔

PR #239, opened 2017-03-28
fix type reference on childContainer's resolve

rcfrias:
The intention of the code is to demonstrate, the childContainer can resolve a Type registered originally for its parentContainer.


jakubvano:
Thanks for catching this! It was probably missed during migration to Swift3 naming conventions.

Issue #228, opened 2017-03-14
Assembler init throws


ManWithBear:
Hello, can you please describe, why this method throws? I checked sources, and here no functions that can throw inside this init.
`public init(assemblies: [Assembly], parentAssembler: Assembler?) throws`

jakubvano:
@ManWithBear Thanks for pointing this out!
Indeed it seems redundant. There is also an [unnecessary optional](https://github.com/Swinject/Swinject/commit/5f6b3d1f6e37b35f9f1394f12d2419c864603c2c#commitcomment-21481767), as pointed out by @tomassliz.

I think we should improve this API. However, it would technically be a breaking API change, so we might need to tread carefully here. @yoichitgy what do you think? Maybe quick improvement might be more important here, given how (in most cases) negligible impact these changes would have on existing code.

yoichitgy:
Thanks @ManWithBear and @tomassliz for pointing out the issues.

Swinject v1 had dependency of `Assember` on [SwinjectPropertyLoader](https://github.com/Swinject/SwinjectPropertyLoader). When I removed the dependency from `Assember`, I forgot taking care of `throws` and the optional argument.

I agree @jakubvano with the quick improvement in Swinject v2. I think we can keep the existing API (initializers) with `@available(*, deprecated)` annotation to avoid breaking changes.

ManWithBear:
Can I help with that?

yoichitgy:
Thanks @ManWithBear😃 Yes, it's very helpful if you make a pull request👍👍👍

ManWithBear:
Thanks, I will try to find time today.

ManWithBear:
Hm, it's kinda weird.
From Swift Programming Language:
> The `throws` keyword is part of a function’s type, and nonthrowing functions are subtypes of throwing functions. As a result, you can use a nonthrowing function in the same places as a throwing one. 

So removing `throws` keyword will not break API and should be fully compatible with previous versions.
But statements with `guard` like:
```swift
guard let childAssembler = try? Assembler(assemblies: [], parentAssembler: assembler) else {
    return
}
```
Produce "Segmentation fault: 11"
Right now it's only issue.

jakubvano:
Even if compiler wasn't buggy in this case, it would still produce warning about no throwing functions occuring within `try` expression.
With your PR it produces deprecation warning, but it is much more clear what change caused it.

jakubvano:
Thanks again for your work :)

PR #241, opened 2017-03-29
Remove redundant throws

ManWithBear:
Provide fixes described in #228


yoichitgy:
Thanks @ManWithBear for fixing the problem👍👍👍

PR #246, opened 2017-04-14
Use the first argument label of Assembler

yoichitgy:
@ManWithBear @jakubvano, I would like to discuss one thing before we release Swinject v2.1.0. We fixed the initializers of Assembler in #241, and we dropped the first argument like this with `_`:

```swift
init(_ assemblies: [Assembly], parent: Assembler?)
```

IMO, using the first argument label `assemblies` is a proper way because the initializer of Assembler is not for type conversion.
https://swift.org/documentation/api-design-guidelines/#argument-labels

What do you guys think about this?

I'll release v2.1.0 after discussing this (merging the pull request or just closing).

### DIFF
--- a/Tests/SwinjectTests/AssemblerSpec.swift
+++ b/Tests/SwinjectTests/AssemblerSpec.swift
@@ -16,7 +16,7 @@ class AssemblerSpec: QuickSpec {
         
         describe("Assembler basic init") {
             it("can assembly a single container") {
-                let assembler = Assembler([
+                let assembler = Assembler(assemblies: [
                     AnimalAssembly()
                 ])
                 let cat = assembler.resolver.resolve(Animal.self)
@@ -35,7 +35,7 @@ class AssemblerSpec: QuickSpec {
             }
             
             it("can assembly a container with nil parent and assemblies") {
-                let assembler = Assembler([
+                let assembler = Assembler(assemblies: [
                     AnimalAssembly()
                 ], parent: nil)
                 let cat = assembler.resolver.resolve(Animal.self)
@@ -47,7 +47,7 @@ class AssemblerSpec: QuickSpec {
             }
             
             it("can assembly a multiple container") {
-                let assembler = Assembler([
+                let assembler = Assembler(assemblies: [
                     AnimalAssembly(),
                     FoodAssembly()
                 ])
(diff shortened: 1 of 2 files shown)


ManWithBear:
Are you sure that it will work?
Deprecated method will generate:
```swift
public convenience init(assemblies: [Assembly], container: Container? = Container()) throws
public convenience init(assemblies: [Assembly]) throws
```
New method:
```swift
public init(assemblies: [Assembly], container: Container = Container())
public init(assemblies: [Assembly])
```

This is reason why I dropped `assemblies`label.

yoichitgy:
Thanks @ManWithBear for writing the reason! I've made sure that `throws` is also a part of a method signature, so we can use `assemblies`.

This one calls the initializer with `throws` (the deprecated one):
`let assembler = try! Assembler(assemblies: [], container: Container())`

This one calls the initializer without `throws` (the newly added one):
`let assembler = Assembler(assemblies: [], container: Container())`

Here is Xcode popup showing the list of initializer signatures:
signature

@ManWithBear, do you think I'm missing anything?

ManWithBear:
Try guard statement as described in https://github.com/Swinject/Swinject/issues/228#issuecomment-290191033

yoichitgy:
Thanks @ManWithBear for the comment. I couldn't reproduce the segmentation fault in my environment with Xcode 8.2.1 and 8.3.1.

Can you give me example code to reproduce it? Or will you try rebuilding the project to check whether the segmentation fault can be resolved?

Tests/SwinjectTests/AssemblerSpec.swift
```diff
@@ -258,7 +258,7 @@ class AssemblerSpec: QuickSpec {
             
             it("can't give entities to parent") {
                 let assembler = Assembler()
-                let childAssembler = Assembler([
+                let childAssembler = Assembler(assemblies: [
```
ManWithBear: If we change this expression to:
```swift
guard let childAssembler = try? Assembler(assemblies: [
    AnimalAssembly()
], parent: assembler) else { return }
```
I will get segmentation fault.
Xcode Version 8.3.1 (8E1000a)

ManWithBear:
@yoichitgy

yoichitgy:
Thanks @ManWithBear😃 I reproduced the segmentation fault with Xcode 8.2.1 and 8.3.1.

Let's add the argument label at Swinject v3 when the obsoleted initializers are removed. Not only this part, but also I want to modify APIs to follow Swift 3 API guidelines😉

PR #248, opened 2017-04-16
Add remark to the Container initializer taking a closure

### LINKED ISSUE #97: Compiling large closures when initializing containers increases build time
The readme gives 

```
    let container = Container() { c in
        c.register(AnimalType.self) { _ in Cat(name: "Mimi") }
        c.register(PersonType.self) { r in
            PetOwner(pet: r.resolve(AnimalType.self)!)
        }
        c.register(PersonViewController.self) { r in
            let controller = PersonViewController()
            controller.person = r.resolve(PersonType.self)
            return controller
        }
    }
```

as an example how to use the library.

However, doing this with larger numbers of dependencies noticeably slows down compilation, as the closure seems to be parsed every time the appModule is used (e.g. to resolve).
In our app, this amounted to around 42 seconds. (https://gist.github.com/Lutzifer/64c00ddf839b866a81274496fa48de04 contains the script I used to measure)

Making an extension on Container like this:

```
extension Container {
   func registerModules() {
        self.register(AnimalType.self) { _ in Cat(name: "Mimi") }
        self.register(PersonType.self) { r in
            PetOwner(pet: r.resolve(AnimalType.self)!)
        }
        self.register(PersonViewController.self) { r in
            let controller = PersonViewController()
            controller.person = r.resolve(PersonType.self)
            return controller
        }

  }   
}
```

and calling it as 

```
    let container = Container() { c in
       c.registerModules()
    }
```

solved this for us, as the closure does not need to be parsed.

Might be worth noting in the Readme.

yoichitgy:
Resolves #97.

- Update README to replace the usage of the initializer.
- Add remark to the initializer talking a closure.

@Lutzifer, do you think my changes can help users to avoid seeing the compile speed problem?

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,17 +211,19 @@ Typically services are registered to a container in `AppDelegate` if you do not 
 ```swift
 class AppDelegate: UIResponder, UIApplicationDelegate {
     var window: UIWindow?
-    let container = Container() { c in
-        c.register(Animal.self) { _ in Cat(name: "Mimi") }
-        c.register(Person.self) { r in
+    let container: Container = {
+        let container = Container()
+        container.register(Animal.self) { _ in Cat(name: "Mimi") }
+        container.register(Person.self) { r in
             PetOwner(pet: r.resolve(Animal.self)!)
         }
-        c.register(PersonViewController.self) { r in
+        container.register(PersonViewController.self) { r in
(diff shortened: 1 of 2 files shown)


yoichitgy:
Thanks @Lutzifer!

PR #234, opened 2017-03-24
Generics improvement. Simplified 'resolve' calls.

dmsl1805:

PR #250, opened 2017-04-23
Set default parameter to `resolve` method in Container

chuganzy:
so that make it enable to resolve without type specification like below.

```swift
container.register(Animal.self) { _ in Cat() }
let animal: Animal = container.resolve()!
```


yoichitgy:
Thanks @chuganzy for the pull request, but unfortunately it is a duplicate of #234 which I didn't merge😢  I should have listened to you carefully before you made the pull request, sorry🙇

PR #252, opened 2017-04-27
Update for Xcode8.3

yoichitgy:
- Updated a unit test for `description` property because printing doubled tuple was fixed by Swift 3.1.
- Updated Quick and Nimble.
- Updated CI configuration.


tkohout:
Interesting, the doubled tuple might also affect autoregistration type warnings since we use description to parse the type from it. I will check it in few weeks when I come from vacation

yoichitgy:
Thanks @tkohout for reviewing and merging😀😃😆

PR #257, opened 2017-05-25
Add tests for the weak object scope

mpdifran:


jakubvano:
Test coverage of weak object scope is currently done by `ContainerSpec_CustomScope` and `WeakStorageSpec`. As object scope behaviour is now implemented outside of the `Container`, we should IMO also test it separately - in general more tests don't hurt, but if they are redundant they may be a potential obstacle to further extendability.

mpdifran:
I couldn't find tests for the `weak` object scope in `ContainerSpec_CustomScope`. It looks like it only uses the `custom` scope. Unless I'm missing something?

jakubvano:
`ContainerSpec_CustomScope` is testing whether `Container` works properly with **any** object scope, which includes built-in object scopes. I guess that naming it `_CustomScope` instead of `_ObjectScope` may be confusing - it was implemented during refactor of object scope behaviour, and `.weak` object scope was added only later.

mpdifran:
Right, it tests whether the container works with any object scope, but I think these new tests still add value. They're essentially testing that the `.weak` object scope functions properly (i.e. that the container doesn't maintain a strong reference to the object).

jakubvano:
You are right, I can see there is an additional value in these tests 👍

yoichitgy:
👍👍👍

Issue #51, opened 2016-01-12
Pass assembler while assembling containers


Nikita2k:
This is not an issue, but a suggestion/idea.

Assume I have a `RootController` and `RootRouter`. The last one is responsible for presenting (modal/push, animation, etc) other controllers. I want to present `AnimalController` (which has it's own dependencies).

I have the following assemblies diagram:
- RootAssembly
  - RootController AssemblyType
  - Services AssemblyType (api, db, etc)
- AnimalAssembly.

After #50 I can make animal assembly with parent assembly (I really need to access db from `AnimalController`).

So here is the problem:

```
class RootControllerContainer: AssemblyType {

    func assemble(container: Container) {

        container.register(RootController.self) { r in
            RootController(router: r.resolve(RootRouter.self)!)
        }

        container.register(RootRouter.self) { _ in
            RootRouter()
        }
    }
}
```

Now I'd like to create a `AnimalController` inside `RootRouter.swift`:

```
func showDogDetails(dog: Dog) {
    let animalController = ???
    rootController.navigationController.pushViewController(animalController, animated: true)
}
```

So the question is how do I construct `AnimalAssembly` with `RootAssembly`? When we were working with `Containers` I was able to create new container with given parent.

So I suggest:
Pass assembler reference in `assemble` func, so I can register components, that is dependant on this assembler.

yoichitgy:
> Now I'd like to create a AnimalController inside RootRouter.swift:

I think I understood from the beginning to here with the `showDogDetails` function:

> So the question is how do I construct AnimalAssembly with RootAssembly? When we were working with Containers I was able to create new container with given parent.

But here I just didn't understand the cause of the problem that you could resolve `AnimalController` with simple `Containers` but couldn't with `Assemblies`.:

> So I suggest:
> Pass assembler reference in assemble func, so I can register components, that is dependant on this assembler.

Also, I just didn't get how to solve the problem with the `Assembler` reference passed to `assemble` function of `AssemblyType`.

Would you explain the question and suggestion parts more? (Is the application in VIPER Architecture? Maybe because I don't know typical DI in the architecture, I couldn't catch the context of your idea.)

I guess you can resolve dependencies if you keep references of `Containers` or `Assemblers` in `RootRouter.swift`. (Just my guess)

Nikita2k:
@yoichitgy Yes. I'm using VIPER and I'm talking about that part: 
[...]

Issue #12, opened 2015-10-09
Support implicit instantiation of UIWindow and its root view controller from "Main" storyboard


yoichitgy:
Support implicit instantiation of UIWindow and its root view controller from "Main" storyboard.

If the feature is supported, the following cases are perfectly resolved.

- Issue #11
- [StackOverflow question](http://stackoverflow.com/questions/33010757/using-swinject-to-inject-a-model-class-instance-into-a-view-class-instance-in-sw)

yoichitgy:
Implemented as a new feature for version 0.3.

josephbeuysmum:
Nice Yoichi. Will you be adding a code example?

yoichitgy:
Documented [here](https://github.com/Swinject/Swinject/blob/master/Documentation/Storyboard.md) :)

josephbeuysmum:
Very nice Yoichi, thanks.

PR #258, opened 2017-05-27
Include xcconfig files in repo, instead of relying on them externally

mpdifran:
I believe there's an issue using `SwinjectStoryboard` (at master where the xcconfigs are included in the repo) with `Swinject` (where the current release does not have the configs included in the repo) through Carthage. Something gets messed up where the project is expecting them at one location, but they're actually in the `Carthage/Checkouts` directory.

This should fix it.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,6 +4,3 @@
 [submodule "Carthage/Checkouts/Quick"]
 	path = Carthage/Checkouts/Quick
 	url = https://github.com/Quick/Quick.git
-[submodule "Carthage/Checkouts/xcconfigs"]
-	path = Carthage/Checkouts/xcconfigs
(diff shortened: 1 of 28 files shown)


jakubvano:
👍 this looks OK to me. We probably want to remove other submodules as well, and integrate Quick / Nimble as frameworks built by carthage instead of subprojects, analogously to [Swinject #51](https://github.com/Swinject/SwinjectStoryboard/pull/51) and [SwinjectAutoregistration #12](https://github.com/Swinject/SwinjectAutoregistration/issues/12).

mpdifran:
I'll make another PR for removing submodules tonight.

yoichitgy:
Thanks🙇

PR #259, opened 2017-05-29
Remove Quick and Nimble submodules, and use pre-built frameworks instead

mpdifran:
Add Carthage/Checkout to .gitignore.
Update .travis.yml to run carthage bootstrap instead of git submodule update.
Copy Quick and Nimble frameworks into the test bundle.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,7 +48,7 @@ Packages/
 # Carthage
 #
 # Add this line if you want to avoid checking in source code from Carthage dependencies.
-# Carthage/Checkouts
+Carthage/Checkouts
 
 Carthage/Build
 
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,1 @@
-[submodule "Carthage/Checkouts/Nimble"]
-	path = Carthage/Checkouts/Nimble
-	url = https://github.com/Quick/Nimble.git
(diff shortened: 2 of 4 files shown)


jakubvano:
Thanks for the PR 👍 
Looks like `carthage bootstrap` step is missing from `.travis.yml` config, so travis is unable to find linked frameworks.
Also, carthage frameworks need to be embeded in the test bundle, otherwise it fails with
>The bundle “SwinjectTests” couldn’t be loaded because it is damaged or missing necessary resources. Try reinstalling the bundle.

mpdifran:
Ok, I'll add the `carthage bootstrap` step to `travis.yml`. I'll make sure they get embedded too.

mpdifran:
Can we run the tests again?

jakubvano:
I restarted the tests, but it they are failing in the linux environment - looks like `carthage bootstrap` should not be run there. For reference see [.travis.yml](https://github.com/Swinject/SwinjectAutoregistration/blob/migrate-to-prebuilt-frameworks/.travis.yml) in SwinjectAutoregistration, which [runs sucessfully](https://travis-ci.org/Swinject/SwinjectAutoregistration/builds/200658970) on linux env.

mpdifran:
Looks like Linux is passing now. I think that one build just needs to be run again.

PR #262, opened 2017-06-06
Fixed segmentation crash on Xcode 9

MadsBogeskov:
The problem was related to sending the `init` method on the storage factories in as a function pointer.

### DIFF
--- a/.Package.test.swift
+++ b/.Package.test.swift
@@ -4,6 +4,6 @@ let package = Package(
     name: "Swinject",
     dependencies: [
         .Package(url: "https://github.com/Quick/Quick", majorVersion: 1),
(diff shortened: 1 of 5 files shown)


mpdifran:
Did you log a radar as well? I see no reason why we couldn't have left it as is (since this seems like a Swift compiler bug).

MadsBogeskov:
Yes: http://www.openradar.me/radar?id=5039608677007360

yoichitgy:
Thanks @MadsBogeskov for fixing the issue👍👍👍 I merged the pull request to `swift4` branch for now because I expect the compiler will fix the issue before Swift 4.0 is released👏

Issue #266, opened 2017-06-10
Recursive `resolve()` deadlocks with `synchronize()`-d resolver


serges147:
I see that internally the `SynchronizedResolver` class uses container's `SpinLock` (which is `NSLock`) to make locks. Can we change the `NSLock` to `NSRecursiveLock` here?

BTW, here's my use case (simplified of course - I do have concurrency):
```
public class A {
    private let b: B
    init(with r: Resolver) {
        b = r.resolve(B.self)    // this dead locks!
    }
}
private class B {}

let c = Container()
let threadSafeResolver = c.synchronize()

c.register(A.self, factory: { _ in return A(with: threadSafeResolver) })
c.register(B.self, factory: { _ in return B() })

threadSafeResolver.resolve(A.self)
```

serges147:
IMO, current behaviour is if not a bug but serious limitation. Any suggestions how to at least work-around it?

mpdifran:
Why can't you just pass `B` into the constructor? 

i.e:

    public class A {
        private let b: B
        init(with b: B) {
            self.b = b
        }
    }
    private class B {}

    c.register(A.self) { r in 
        let b = r.resolve(B)!
        return A(with: b) 
    }

serges147:
Becase i don't want expose dependency to B. Also this is simle example - in reality I resove 3-5 different services as well in A's init. Definitelly would like to keep all of them as implementation detals.

serges147:
Note, that your code will deadlock as well.

yoichitgy:
Why don't you use a container like this?

```swift
c.register(A.self, factory: { resolver in return A(with: resolver.resolve(B.self)!) })
c.register(B.self, factory: { _ in return B() })
let threadSafeResolver = c.synchronize()
```

serges147:
Let me try your code - I don't see big difference with mine, but may be I'm wrong... Will be back in 15 min

serges147:
Here we are. Check out attached sample project - deadlocks as I thought.
[SwinjectLock.zip](https://github.com/Swinject/Swinject/files/1066089/SwinjectLock.zip)

serges147:
Changing SpinLock.swift#12...

from: `private let lock =  NSLock()`
to: `private let lock =  NSRecursiveLock()`

...immediately fixes my `test__deadlocks_on_second_resolve`

mpdifran:
@yoichitgy I believe your solution is the same as mine.

serges147:
[...]

PR #269, opened 2017-06-12
Fix deadlock when recursive resolving made on "thread safe" resolver

serges147:


yoichitgy:
I think capturing the resolver returned from `container.synchronize()` in `register` closure is just misuse of SynchronizedResolver.

serges147:
Could you please what foundamentially wrong with it? I agree that if you have `resolver` as first parameter of factory callback then you don't need to capture anything else - this as exctly as I did. Then I get wierd crash because of not being threadsafe with resolving, so I added the synchronize() - and guess what, even then you are not ok and Swinject wont help you. So capturing outside `threadSafeResolver` is kinda workaround, and IMO quite ligitimate one.

yoichitgy:
I appreciate you tried to improve Swinject, but please check Issue #266😉

jakubvano:
Original `it wont resolve the same instance` test was not valid use of swinject - not even `NSRecursiveLock` could enable that.
However, enabling nested resolves is possible by using `NSRecursiveLock` - I reimplemented that test in #351.

Thanks for your contribution 👍

jakubvano:
Closing in favor of #353 & #350

Issue #197, opened 2016-11-30
Enable Swiftlint for Tests


jakubvano:
As discussed [here](https://github.com/Swinject/Swinject/pull/196#discussion_r90190293).

jakubvano:
Resolved by #270

PR #270, opened 2017-06-13
Swiftlint cleanup

jakubvano:
This enables swiftlint for tests resolving #197, fixes violations, and enables some of the disabled rules - my reasoning is, that rule should not be globally disabled, unless there is a good reason to do that:

- `trailing_whitespace`: automatic trimming can be enabled in Xcode preferences
- `opening_brace`: this was an issue only for functions with multiple parameters, and can be solved by alternative return type alignment
- `nesting`: violation could be fixed quite seemlessly by extracting the nested type.
- `line_length`: this is IMO very important rule as it - among  other stuff - guarantees good readability in github diff.
- `type_body_length` & `function_body_length`: this is an issue only in spec files due to how QuickSpecs are structured, and should IMO be disabled expilicitly in spec files (this is how Quick templates do it)


yoichitgy:
Thank you for the great improvement👍👍👍

PR #287, opened 2017-08-17
Add RegistryType

matsuokah:
Add RegistryType to specify the registration without String.

## Motivation
In my opinion, specifying the registration from String is little fragile.
And I want to specify it by enum.

So, in my project, I write like below.

```swift
enum RegistrationType: String {
  case A
  case B
  
  var name: String {
    return rawValue
  }
}

let registration = container.resolve(Class.self, name: RegistrationType.A.name)
```

But, I do not want to write the `.name`, because that is a little wordy, then I create this pull request.

Best regards.


matsuokah:
This PR is unstable, I'll fix that.

PR #288, opened 2017-08-17
Add RegistryType

matsuokah:
Add RegistryType to specify the registration without String.

## Motivation
In my opinion, specifying the registration from String is little fragile.
And I want to specify it by enum.

So, in my project, I write like below.

```swift
enum RegistrationType: String {
  case A
  case B
  
  var name: String {
    return rawValue
  }
}

let registration = container.resolve(Class.self, name: RegistrationType.A.name)
```

But, I do not want to write the `.name`, because that is a little wordy, then I create this pull request.

Best regards.

### DIFF
--- a/Sample-iOS.playground/Contents.swift
+++ b/Sample-iOS.playground/Contents.swift
@@ -19,7 +19,7 @@ class Cat: Animal {
     init(name: String?) {
         self.name = name
     }
-    
+
     func sound() -> String {
         return "Meow!"
     }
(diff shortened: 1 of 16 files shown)


yoichitgy:
Thanks @matsuokah for the PR👍 I'll comment soon🙏

jakubvano:
Thanks again for the PR! 👍 

However I think this is an unnecessary extension: If the goal is to get rid of magic-strings, then it can be achieved simply by:

```swift
enum RegistrationType {
    static let A = "A"
    statis let B = "B"
}

container.resolve(Class.self, name: RegistrationType.A)
```
We would like to keep only basic functionality in the main repo, optional features (like this one) are probably better suited for self-standing repo.

jakubvano:
I will close this due to the lack of activity - feel free to reopen it if you have related ideas or have a different viewpoint 😉

PR #298, opened 2017-09-30
Use hashValue from ObjectIdentifier instead of described String.

tarunon:
Hi there,
This pr is super [nits], ObjectIdentifier is usable when you wish hashValue from metatype or class value. I think its better than `String(describing:_)`.
ObjectIdentifier documents is here. https://developer.apple.com/documentation/swift/objectidentifier

If you don't wish to use feature around Reflecting, please close this pr.


yoichitgy:
Thanks @tarunon for your PR with the link to Apple's doc👍👍👍 You gave me better understanding of Swift🙏

PR #300, opened 2017-10-05
Change PersistentStorage to PermanentStorage

jkaan:
PersistentStorage does not exist at the moment. This refers to the
correct class for this object scope.

### DIFF
--- a/Documentation/ObjectScopes.md
+++ b/Documentation/ObjectScopes.md
@@ -39,7 +39,7 @@ Above holds for reference types - value types are not shared in this object scop
 Custom object scopes can be defined like this:
 ```swift
 extension ObjectScope {
-    static let custom = ObjectScope(storageFactory: PersistentStorage.init)
+    static let custom = ObjectScope(storageFactory: PermanentStorage.init)
 }
 ```
 Instances in `.custom` scope will be shared in the same way as in `.container` scope but can be discarded as needed:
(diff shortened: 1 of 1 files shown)


yoichitgy:
Thanks for your PR👍 The doc is an example to create your own custom scope. `.graph` and `.container` scopes uses `PermanentStorage.init`.
https://github.com/Swinject/Swinject/blob/5c84eeffcf5590323258b65145a211361cc03346/Sources/ObjectScope.Standard.swift

To clarify the intention of the document, I think changing the part to `YourCustomStorage.init` looks better. What do you think about it?

jkaan:
I do agree! However at that moment it would make sense to also add a little description on how to create a `YourCustomStorage.init`. Shall I extend this PR to also include a little bit of documentation on that? :)

yoichitgy:
Sure👍 It's very helpful for Swinject users if you enhance the documentation👍👍👍

jkaan:
@yoichitgy Could you take another look at this PR, please? I added a tiny bit of documentation on creating your own implementation of InstanceStorage.

jakubvano:
IMO typical usage of custom object scope would use one of the default `InstanceStorage` implementations, that is why initial documentation didn't go into detail of explaining how to write `YourCustomStorage` (Your example is basically reimplementation of `PermanentStorage`).

I feel like use of custom `InstanceStorage` would have to be very specific and require quite detailed understanding of inner workings of `ObjectScope` system. Thus IMO it is not worth trying to cover it in the docs.

What do you think @jkaan ?

Issue #291, opened 2017-09-08
Make `.container` the default object scope


Herriau:
The fact that the `.graph` scope is assumed by default led me to very confusing results in so far as I did not expect my factories to be called more than once. It took me a while to figure out what was going on, especially because this concept of object scope isn't mentioned anywhere in the README.

It seems to me that `.inObjectScope(.container)` is the behavior that I have come across in most DI frameworks. Should this be the default instead?

Herriau:
* Php-di
* PHP Symfony's container
* InversifyJS

Regardless of the existence of that ObjectScopes page, I feel like if `.graph` is to remain the default, this should be mentioned in big letters on the main README.

olejnjak:
I'd add my bit to this thread. I also feel like `.graph` scope is not what I need in my apps. But also I understand that changing this would semantically big breaking change. 

But my case is a bit different, my ideal default scope would be `.transient`. I use SwinjectAutoregistration and almost every factory I register assumes that every object return is newly created.

My suggestion would be to keep `.graph` as default scope, but allow developers to change it.

yoichitgy:
Thanks @olejnjak for your idea👍

jkaan:
@yoichitgy I could (try to) create a PR to make the default scope part of the container as a configuration option. What do you think?

serges147:
Yup, once I realized that `.graph` is default scope, I had to look at every and each registration place, and explicitly specify either `.container` or `. transient` depending on context.

jakubvano:
Resolved by #304

PR #304, opened 2017-11-02
Add possibility to specify default object scope to container

jkaan:
This allows you to specify the default object scope that the container uses to create the service entries. It is still defaulting to `.graph` to keep the behaviour equal.

This should also solve the issue mentioned here: https://github.com/Swinject/Swinject/issues/291

### DIFF
--- a/Sources/Container.swift
+++ b/Sources/Container.swift
@@ -28,12 +28,14 @@ public final class Container {
     fileprivate let parent: Container? // Used by HierarchyObjectScope
     fileprivate var resolutionDepth = 0
     fileprivate let debugHelper: DebugHelper
+    fileprivate let defaultObjectScope: ObjectScope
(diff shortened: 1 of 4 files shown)


Steven-Wu-ConnectedLab:
Any plans to merge this in?  This is a really helpful feature.

MortyMerr:
Any plans to merge this in?

connected-swu:
Any plans to merge this?

kerry:
Any plans to merge this? @yoichitgy

jakubvano:
Thanks for the PR! 👍 
In addition to this, it would be useful to extend `Assembler` to fully support default scope (as it stands, you cannot specify default scope if creating assembler with parent).

But this is a great addition on its own, assembler support can be implemented separately.

yoichitgy:
Thanks @jkaan for making PR, and @jakubvano for merging it👍

PR #319, opened 2018-02-16
Type forwarding

jakubvano:
This implements API for type forwarding - heavily inspired by https://github.com/AliSoftware/Dip/wiki/type-forwarding.

These changes should be backwards compatible for standard Swinject users, however there are breaking changes in the `_Resolver._resolve` and `Container._register` methods - which might affect Swinject extensions using this methods (SwinjectStoryboard, SwinjectAutoregistration).


yoichitgy:
👍

jakubvano:
I've added the documentation to this API.

@yoichitgy What do you think? Do you have some ideas for the API / implementation improvements? 

As mentioned above, in order to implement this feature I had to change the signature of some public methods - even though they should not be used by regular users of the Swinject. Do you think we can release this with the `2.3.0` in spite of these breaking changes?

yoichitgy:
I think it's acceptable to release it as `2.3.0` if the methods have documentation comments telling they are used only for extensions like SwinjectStoryboard 👍

PR #329, opened 2018-03-06
Make pod as static framework

AnisovAleksey:
Hi to all

Cocoapods have introduced support of static frameworks:
https://github.com/CocoaPods/CocoaPods/pull/6811

As I know it improves app startup time.

### DIFF
--- a/Swinject.podspec
+++ b/Swinject.podspec
@@ -15,5 +15,6 @@ Pod::Spec.new do |s|
   s.osx.deployment_target = '10.10'
   s.watchos.deployment_target = '2.0'
   s.tvos.deployment_target = '9.0'
+  s.static_framework = true
   s.requires_arc = true
 end


jakubvano:
Thanks for the suggestion 👍 Until now I didn't know about this cocoapods feature. 

🤔 I'm afraid though that this might break integration for some current projects using Swinject as dynamic framework. Will have to take a detailed lookinto  how it works, and what potential impacts it may have.

jakubvano:
@AnisovAleksey Do you know of some popular libraries, that already use this feature?

AnisovAleksey:
@jakubvano after a little study, I got that current implementation in CocoaPods is weird. Dynamic framework type will be unavailable and it can affect another projects\libraries. 
Still wait for better support for static frameworks in CocoaPods :(

Issue #303, opened 2017-10-25
Resolve object lazily


AnisovAleksey:
Hello!

In some cases we need to create wrapper like a ViewFactory or something like it. In general it is lazy initialization, so would be nice if Swinject will support lazy initialization.
It can be implemented as struct that contains Closure to resolve object. example of implementation below:

```
// Object will be common if someone have a strong reference
public struct Lazy<Service> {
    private let initClosure: () -> Service?
    private var service: Service?
    
    internal init(initClosure: () -> Service?) {
        self.initClosure = initClosure
    }
    
    public func resolve() -> Service? {
        if service == nil {
            service = initClosure()
        }
        return service
    }
}
```

What you think about it?

jakubvano:
Hello @threat70, thanks for the suggestion! 👍 

I understand your reasoning, and can see something like this being added to the Swinject. Do you have some suggestions how an API might look like / some references to such feature in different DI frameworks?

I see a couple of issues (i.e. how would `ObjectScope.graph` work with lazy instantiation?), but i think it may be doable.

threat70:
Hello @jakubvano.

I think that "lazy" should be a new method of registration, I mean that use something like `container.registerLazy(CustomView.self) { .... }` that will return `Lazy<Service>`, object scope will affect only `Lazy<Service>` object. `Lazy<Service>` store the last resolved instance of `Service` and provide it, so if you have the same `Lazy<Service>` object in other place, you will have the same `Service` at this places.

Also for another case can be implemented object like `Provider<Service>`, that will resolve `Service` for each call. 

I took it from Dagger 2+. ref: https://google.github.io/dagger/api/2.10/dagger/Lazy.html

Thanks

J7mbo:
How about just providing an optional `lazy` argument to the current api?

AnisovAleksey:
@J7mbo what do you actually mean? could you provide sample code?

J7mbo:
@AnisovAleksey I tried something the other day, whereby I inject a closure returning the object instead. 

```
container.register(MyObj.self) { in
    return { MyObj() 
}
```

Then in your object that requires `MyObj`, you type hint for the closure:

```
class MyClass
{
     public var myObj: (() -> (MyObj))?
}
```

So when `MyClass` is initialised (and resolved from the container), only the closure containing `MyObj` is. You don't actually instantiate the object until, somewhere in your code, you call `self.myObj()`.

[...]

Issue #322, opened 2018-02-19
Behaviours


jakubvano:
I've been thinking of adding **Behaviours** (or **Extensions**, name pending 😄 ) feature: 
```swift
public protocol Behavior {
    func container<Service>( // This would be invoked for every registered service
        _ container: Container,
        didRegisterService entry: ServiceEntry<Service>,
        withName name: String?
    )
}
```
You could implement your own behaviours and then add them to the container
```swift
container.addBehaviour(MyBehaviour())
```

This would enable lots of previously requested features with minimal changes to existing codebase, e.g.:

**Optional Forwarding** (with #319, very useful for SwinjectAutoregistration [#28](https://github.com/Swinject/SwinjectAutoregistration/issues/28), #280)
```swift
public struct OptionalForwarding: Behavior {
    public func container<Service>(
        _ container: Container,
        didRegisterService entry: ServiceEntry<Service>,
        withName name: String?
    ) {
        container.forward(Service?.self, name: name, to: entry)
        container.forward(Service!.self, name: name, to: entry)
    }
}
```

**Instance Aggregation** (#271 )
```swift
public class InstanceAggregation<BaseType>: Behavior {
    private let isChild: (Any.Type) -> Bool
    private var factories = [(Resolver) -> BaseType?]()

    public init(isChild: @escaping (Any.Type) -> Bool) {
        self.isChild = isChild
    }

    public func container<Service>(
        _ container: Container,
        didRegisterService entry: ServiceEntry<Service>,
        withName name: String?
    ) {
        guard isChild(Service.self) else { return }

        if factories.isEmpty {
            container.register([BaseType].self) { resolver in self.factories.flatMap { $0(resolver) } }
        }
        factories.append { $0.resolve(Service.self, name: name) as? BaseType }
    }
}

container.addBehavior(InstanceAggregation<Animal>(isChild: { $0 is Animal.Type }))
```

With a bit of further work posiibly even some kind of **Base Class Registration** (#108).

@yoichitgy @tkohout What do you guys think?

jakubvano:
Implemented by #330

Issue #221, opened 2017-02-12
Assembly/Assembler need thread safety


kdubb:
Unless I'm missing something, there is currently no way to call `synchronized` from an `Assembly` and the `resolver` property doesn't apply it.  So, using `Assembly` seems to always be unsafe.

For the moment I've copied `SynchronizedResolver` into our project (the version provided by Swinject is not public) and am wrapping the `Assembly.resolver` with it.

kdubb:
I should say I am using a heavily modified `SynchronizedResolver` that uses a serial `DispatchQueue` as a lock and only requires a `Resolver`.

yoichitgy:
I agree👍 As we talked in #139, I also want to enhance concurrency model. I'll try to implement concurrency for Assembly/Assembler. If you can show me your modified code, it's very helpful!

geraldeersteling:
How is this coming along? I'm also worried about thread safety.

Right now I'm creating my `Assembler` with the SwinjectStoryboard's `defaultContainer`. I'm thinking of exposing the `defaultContainer`'s threadsafe view (`synchronize`) instead of the `Assembler`'s `resolver`. Thoughts?

sprynmr:
So what is the mechanism for exposing a thread safe resolver from the Assembler?

yonivav:
@sprynmr `let synchronizedResolver = (assembler.resolver as! Container).synchronized()`
from [(https://stackoverflow.com/questions/49687342/crash-container-swift-line-299-container-resolvea-b-entry-serviceentrypro)]

devxoul:
I've created a patch for the issuse: #445. I would really appreciate if somebody could review this! Thanks :)

bryanwagner:
The PR submitted above #445 appears to fix the issue. As a workaround, I'm using the following extension function until the PR is merged:
```
extension Assembler {
    public var synchronizedResolver: Resolver {
        let internalResolver = resolver
        if let it = internalResolver as? Container {
            return it.synchronize()
        } else {
            return internalResolver
        }
    }
}
```

ursusursus:
Is this library dead?

AndreiTich:
I would love for this fix to be officially merged. Any updates?

Issue #139, opened 2016-08-31
Simplify thread safety model


yoichitgy:
Currently a thread-safe container is created by `synchronize` method wrapping the receiver container in `SynchronizedResolver`. It has turned out that this model is difficult to maintain to keep consistency of `ResolverType` with `Container` and `SynchronizedResolver`.

It looks good to pass a dispatch queue to `Container` initializers to use in `resolve` methods with `dispatch_sync`. If no dispatch queue is passed to the initializers, the container is not thread safe (but its performance is better).

Another motivation for the change on the thread safety model is OSSpinLock was previously used for the best performance, but it has been replaced with NSLock for its problem (#116). Using a dispatch queue is better in performance than NSLock, and looks good for usability of Swinject.

yoichitgy:
I wrote like above, but maybe we might need something like a recursive lock??? though dispatch queue is preferable.

kdubb:
@yoichitgy It seems like concurrency might better be handled at the `ObjectScope` level rather than the `Container` level. After all, it's the scope the determines if it needs thread-safety. 

It seems to me...
`container` needs a simple lock to ensure it's only created once (e.g. like `dispatch_once`.
`transient` shouldn't need any specific locks
`weak` only needs an atomic test (handled by swift)
`graph` needs a recursive lock

I'm not quite sure how this is all implemented so maybe it isn't as easy as I've laid out. From an outside perspective the locking seems like it should act in this way though.  Am I missing something?

yoichitgy:
@kdubb thank you for your great idea😃 I was thinking passing a DispatchQueue to a Container might be good for concurrency, but handling at an ObjectScope level looks nice because @jakubvano improved extendablity of ObjectScope.

One thing we should care about is complexity in SwinjectStoryboard, SwinjectPropertyLoader and SwinjectAutoregistration. So far, for them, it looks better to handle concurrency in ObjectScope.

I'll investigate ObjectScope implementation for concurrency. If you got a good idea for implementation, please write here. Always welcome 😃

jakubvano:
Handling concurrency at `ObjectScope` level is interesting idea, but I don't think it is the right place - in my mind `ObjectScope` is responsible for instance **persistency**, while `Container` is responsible for instance **creation**. Also, implementation would be complicated, as currently `InstanceStorage` has separate methods for storing and retrieving an instance, and ensuring thread safety requires synchronisation of both operations.

jakubvano:
Closing due to inactivity

PR #331, opened 2018-03-08
Lazy injection

jakubvano:
My take on implementing the Lazy injection (#303). It definitely needs code cleaning, synchronicity considerations, and maybe some optimizations / api improvements.

Biggest change in the existing codebase is object graph caching, which is necessary to implement Lazy injection, as @J7mbo [pointed out](https://github.com/Swinject/Swinject/issues/303#issuecomment-370836649).

The idea is to generate an unique identifier for each graph resolution, and provide that information to  the`InstanceStorage` when retrieving / storing instances. Then `GraphStorage` is caching instances resolved in specific graphs, to return them if graph is restored. I'm keeping only weak references - this should be enough for purposes of graph caching, AFAICS has negligible memory impact and shouldn't lead to zombie instances.


jakubvano:
While trying to make this feature more robust (e.g. properly work for registrations with arguments) I've converted to the solution, which basically makes lazy resolution implicit for all services with (AFAICS) minimal impact on performance:

There is a `InstanceWrapper` protocol, which is explicitly checked in the case when default resolution mechanism fails (i.e. there is no entry for `InstanceWrapper` type). When this happens, instance of the wrapper is created and transparently returned from `resolve` method.

As this additional resolution attempt is performed only when default mechanism fails, performance impacts on existing code should be minimal. 

And as a bonus, this `InstanceWrapper` concept should easily enable optional forwarding (much more efficiently then [this](#322) approach) - all one needs to do is:

```swift
extension Optional: InstanceWrapper {
    static var wrappedType: Any.Type { return Wrapped.self }
    
    init(inContainer container: Container, withInstanceFactory factory: @escaping () -> Any?) {
        self = factory() as? Wrapped
    }
}
```

jakubvano:
I've been trying to make the `Lazy` feature thread safe, but it does not seem to be possible in the current model - i.e. i see no way of synchronizing instance resolution inside `Lazy.instance` getter.

AnisovAleksey:
@jakubvano wow, your idea about `InstanceWrapper` is very good! 👍 
What problems still exist? (I mean why it is still WIP? :) )

jakubvano:
Currently there are two areas in which I'm not completely sure how to proceed:

1. **Thread safety**  - as mentioned above, I see no way of making `InstanceWrapper`s thread safe - it would require redesign of thread safety model (#139). However, I'm leaning towards releasing the feature despite of that (e.g. `Assembler`s do not support thread safety yet #221).

2. **Provider circular dependency** - `Lazy` circular dependencies work as expected because of graph caching. However, this mechanism cannot be used for `Provider` dependencies, as it would result in always returning the same instance.
My dilemma here is whether we should make circular dependencies work with `Provider` as well, or is it something that user would not need - One example would be something like this:

```swift
class Controller: WorkDelegate {
    private let worker: Provider<Worker>
}

extension Controller: WorkerDelegate {
    func workCompleted() {  }
}

class Worker {
   weak var delegate: WorkerDelegate?
}
```
Here I might expect, that every provided worker would have have the same instance of `Controller` as delegate.

Would definitely appreciate any input on this :)

jakubvano:
After further consideration, I came to the conclusion that expecting the `Provider` circular dependencies to work in `.graph` goes - in a way - against the main idea behind the `Provider`, would probably be useful only in very niche situations, and thus it is not worth to severly complicate the implementation of graph caching.

PR #332, opened 2018-03-11
Fix README for swift 4

acevif:
@objc attribute is mandatory for SwinjectStoryboard's setup() class
function in swift 4.
see https://github.com/Swinject/SwinjectStoryboard#uiwindow-and-root-view-controller-instantiation
for more details.


jakubvano:
Good job catching that 👍 Thanks for your contribution 😄

PR #338, opened 2018-03-30
Swift 4.1

gilroykilroy:
Changed travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode9.2
+osx_image: xcode9.3beta
 env:
   global:
     - LC_CTYPE=en_US.UTF-8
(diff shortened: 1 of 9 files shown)


gilroykilroy:
Do note SwinjectStoryboard will have to be modified to point to this version once integrated.

.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode9.2
+osx_image: xcode9.3beta
```
pbyorke: Does this mean the change will only affect the beta version of Xcode 9.3? I'm on the real deal: Version 9.3 (9E145)

gilroykilroy:
Closing this one since #342 contains all I did and then some. The travis change was only for the automated test tool which hasn't been updated. Also neither my change nor 342 is backwards compatible to swift 4.0. I suppose those changes should have been wrapped in a #if swift(>=4.1) for those of you still pre-XCode 9.3 and for travis.

Anyway, my repos (gilroykilroy/swinject and gilroykilroy/swinjectstoryboard) are public and have my changes. So if you are on Xcode 9.3 and use carthage feel free to point to those until the changes are integrated into the real repo.

Issue #280, opened 2017-07-29
Resolve optionals and implicitly unwrapped optionals


tkohout:
I would like to add new feature to Swinject. The following code now fails to resolve:

```
container.register(Service.self) { _ in Service() }
        
let vc1 = container.resolve(Service?.self) // nil
let vc2 = container.resolve(Service!.self) //nil
```

Even though this case rarely bothers regular Swinject users it does matter when you use SwinjectAutoregistration. Any service with optional or implictly unwrapped dependency have to be registered by hand and it brings inconsistency to the api. 

I do think the `Service`, `Service?` and `Service!` are logically the same and should resolve in the same service. 

I can't implement this in the SwinjectAutoregistration due to generics limitation but it can be done in the Swinject itself. I took inspiration in [DIP](https://github.com/AliSoftware/Dip/wiki/resolving-optionals) where the type-forwarding is used for the same purpose. For our case something like this could be implemented in `ServiceEntry`:

```
let resolvableTypes: [Any.Type] = [((Resolver)->Service?).self, ((Resolver)->Service!).self]

func canResolve(type: Any.Type) -> Bool {
        return resolvableTypes.contains(where: { $0 == type})
}
```

and the entries could be found after the current lookup fails with something like this:

```
let entry = services.filter ({ $1.canResolve(type: key.factoryType) }).first?.value
```

I would like to get a feedback from @yoichitgy or @jakubvano  whether this is something that could be added or whether it might break something.

yoichitgy:
What is the motivation to add the feature? What usecases do you expect? Technically we can add the feature, but it introduces complexity to our codebase. IMO it's good to add the feature only if the cost can be paid enough💸 (Of course, the feature and implementation are technically interesting😉

yoichitgy:
I think type-forwarding is worth implementing because we had requests/questions from time to time (#28, #227, #247), but the optional resolution is not highly prioritized. The policy of Swinject project is making a reliable DI container in **production**.

In case of the `APIService` example, injecting an empty implementation to non-optional argument makes `APIService` simpler because you don't have to unwrap `logger` in `APIService`. IMO, we should recommend the empty implementation.

```swift
class APIService {
   init(logger: Logger)
}

class EmptyLogger: Logger {
    func log(_ message: String) {
        // Do nothing as empty implementation.
    }
}
```

---

[...]

PR #342, opened 2018-04-02
Update the project for Xcode 9.3 and Swift 4.1

Vkt0r:
Hi guys, below are the list of changes in the project:

• Update the **.travis.yml** to run on the Xcode9.3beta version.
• Replace the deprecated `flatMap` operator in favor of `compactMap`.
• Fix a lint warning with nesting type.
• Remove unnecessary implementation of the `InstanceWrapper ` by the `ImplicitlyUnwrappedOptional`.
• Fix an issue causing the compiler was not inferring the types for the `+` operator correctly declaring the types explicitly for the closure.
• Disable the `line_length` rule in the **ContainerSpec.Arguments** file to avoid the warnings.

Regarding the `+` operator problem with the Swift compiler I solve it in one of the several ways, we can solve it (explicitly declaring the types of the arguments in the closure, as the most of the times in type inference problems of the compiler). The `joined` function can be another solution as pointed by Rob Napier [here](https://stackoverflow.com/questions/39490808/how-to-merge-multiple-arrays-without-slowing-the-compiler-down/39511195#39511195).

### DIFF
--- a/Tests/SwinjectTests/ContainerSpec.TypeForwarding.swift
+++ b/Tests/SwinjectTests/ContainerSpec.TypeForwarding.swift
@@ -129,27 +129,27 @@ class ContainerSpec_TypeForwarding: QuickSpec {
             it("resolves optional when wrapped type is registered") {
                 container.register(Dog.self) { _ in Dog() }
                 let optionalDog = container.resolve(Dog?.self)
-                let unwrappedDog = container.resolve(Dog!.self)
+                let unwrappedDog = container.resolve(Dog?.self)
                 expect(optionalDog ?? nil).notTo(beNil())
                 expect(unwrappedDog ?? nil).notTo(beNil())
             }
             it("resolves optional to nil when wrapped type is not registered") {
                 let optionalDog = container.resolve(Dog?.self)
-                let unwrappedDog = container.resolve(Dog!.self)
+                let unwrappedDog = container.resolve(Dog?.self)
                 expect(optionalDog).notTo(beNil())
                 expect(unwrappedDog).notTo(beNil())
             }
             it("resolves optional with name") {
                 container.register(Dog.self, name: "Hachi") { _ in Dog() }
                 let optionalDog = container.resolve(Dog?.self, name: "Hachi")
-                let unwrappedDog = container.resolve(Dog!.self, name: "Hachi")
+                let unwrappedDog = container.resolve(Dog?.self, name: "Hachi")
                 expect(optionalDog ?? nil).notTo(beNil())
                 expect(unwrappedDog ?? nil).notTo(beNil())
             }
             it("resolves optional to nil with wrong name") {
                 container.register(Dog.self, name: "Hachi") { _ in Dog() }
                 let optionalDog = container.resolve(Dog?.self, name: "Mimi")
-                let unwrappedDog = container.resolve(Dog!.self, name: "Mimi")
+                let unwrappedDog = container.resolve(Dog?.self, name: "Mimi")
                 expect(optionalDog ?? nil).to(beNil())
                 expect(unwrappedDog ?? nil).to(beNil())
                 expect(optionalDog).notTo(beNil())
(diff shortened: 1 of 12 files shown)


Tests/SwinjectTests/ContainerSpec.TypeForwarding.swift
```diff
@@ -129,27 +129,27 @@ class ContainerSpec_TypeForwarding: QuickSpec {
             it("resolves optional when wrapped type is registered") {
                 container.register(Dog.self) { _ in Dog() }
                 let optionalDog = container.resolve(Dog?.self)
-                let unwrappedDog = container.resolve(Dog!.self)
+                let unwrappedDog = container.resolve(Dog?.self)
```
jakubvano: Ideally we do not want to lose this functionality for implicitly unwrapped optionals. Do these tests fail without `ImplicitlyUnwrappedOptional: InstanceWrapper {}` conformance?

jakubvano:
Thanks for the PR! 👍 
In order to properly run the tests on Linux, we need to update `.swift-version` file.

One issue with this PR is, that it removes `InstanceWrapper` functionality for `ImplicitlyUnwrappedOptionals` - which we ideally want to keep - if there is no way to declare conformance to `InstanceWrapper`, we might need to find other ways to make that work..

Vkt0r:
@jakubvano You're welcome !!! Regarding the SPM failing test, I was able to enable `flatMap` for swift version `<4.1`. The job that is failing and I don't know exactly why? because TravisCI is not giving too much information about it is the `JOB="XCODE" DEST="OS=8.4,name=iPhone 5" SCHEME="Swinject-iOS" SDK="iphonesimulator" ACTION="test"`. If you could take a look it could be great!

Tests/SwinjectTests/ContainerSpec.TypeForwarding.swift
Vkt0r: @jakubvano I removed the `InstanceWrapper` functionality for `ImplicitlyUnwrappedOptionals` because the compiler was complaining about the conformance as the type `ImplicitlyUnwrappedOptionals` with the following message:

```bash
Using 'ImplicitlyUnwrappedOptional' in this location is deprecated and will be removed in a future release; consider changing this to 'Optional' instead
```

And we already have the conformance for `Optional` above that part of the code:

```swift
extension Optional: InstanceWrapper {
    static var wrappedType: Any.Type { return Wrapped.self }

    init?(inContainer container: Container, withInstanceFactory factory: (() -> Any?)?) {
        self = factory?() as? Wrapped
    }
}
```

So please tell me what do you want to do here 🤔.

jakubvano: I see, after reading through the [0054-abolish-iuo.md](https://github.com/apple/swift-evolution/blob/master/proposals/0054-abolish-iuo.md) it seems that implicitly unwrapped optionals will be completely removed in future swift release.

Curently it is not yet the case however and as a result of this change implicitly unwrapped optional dependencies will not be resolved (this will probably only affect SwinjectAutoregistration, see #280).

I don't see an easy way of circumventing this though, and given that it will be fixed in future swift release makes it probably not worth the effort.

jakubvano: So your change is OK, although there is no point in duplicating the tests - we can remove all the`unwrappedDog` lines.

jakubvano:
I can reproduce the issue with failing job locally - tests fail on ios 8.x simulator. However, I can reproduce this issue on empty project with default tests, thus it seems it is some kinf of a bug introduced by Xcode 9.3.

I'd say we can (temporarily) disable this job in `travis.yml`

Vkt0r:
@jakubvano Yes same here. I was reading some thread about this error in TravisCI and I didn't find anything useful, at some point it disappears. I'm afraid that was introduced in Xcode 9.3 😢. Yes, I think we can temporarily disable the job in `travis.yml`. Say the word and I can update the PR with this change !!

Tests/SwinjectTests/ContainerSpec.TypeForwarding.swift
Vkt0r: Sounds good let me take a look at that tests

Vkt0r:
@jakubvano I remove the references to `unwrappedDog` from the test and disable temporarily the job for iOS 8.x in `travis.yml`.

jakubvano:
👍 looks good! Thanks again for your contributiuon 😄

Vkt0r:
@jakubvano You're welcome 👍

Issue #99, opened 2016-05-16
Throwing resolve methods


jakubvano:
As there are plans for `2.0`, I thought I would throw in this idea:

When trying to unit test my assemblies, I realized it would be nice to "optionaly" resolve instances, i.e. having `resolve()` return `nil` when type cannot be resolved. Currently it returns `nil` only if there are no registrations for given type, but there is no way to return `nil` if there are no registrations for type's dependencies.

One way to solve this, would be changing `resolve()` behaviour so that instead of returning nil when there is no registration it would throw an error.

```
func resolve<T>(serviceType: T.Type) throws -> T
```

Then in registration closures you could write

```
c.register(PersonType.self) { r in
    PetOwner(pet: try r.resolve(AnimalType.self))
 }
```

I realise that in most cases being able to optionally resolve types is not needed, but I was wondering what do others think about such a change.

Thoughts?

yoichitgy:
Thanks for your idea to improve Swinject 😃

I think `Container` is a kind of `Dictionary` returning nil is enough for non-existing key. Repeating `try` is redundant. I want to keep it simple.

jakubvano:
Ok, I agree that it would be redundant for most usecases.

yoichitgy:
If Swift language doesn't require `try` for every throwing function, the API you proposed is great😉

jakubvano:
Well, if you have multiple throwing functions in one expression, you can move try an the start of expression. i.e:

```
c.register(SomeType.self) { r in
    try SomeImpl(
        dep1: r.resolve(Dependency1),
        dep2: r.resolve(Dependency2)
    )
 }
```
Instead of

c.register(SomeType.self) { r in
    SomeImpl(
        dep1: try r.resolve(Dependency1),
        dep2: try r.resolve(Dependency2)
    )
 }
```

PR #348, opened 2018-05-15
Added throws resolve

Lumyk:


jakubvano:
Thanks for the PR! 👍 

What is the reasoning behind this functionality? Can you describe why it is necessary / beneficial? How would this change in API impact current users?

Lumyk:
Added ability to make throws resolve

```swift
let container = Container()
container.register(Animal.self) { _ in Cat(name: "Mimi") }
container.register(Person.self) { r in
    PetOwner(pet: r.resolve(Animal.self)!)
}

do {
  let person = try container.throwsResolve(Person.self)
   person.play() // prints "I'm playing with Mimi."
} catch let error {
  // get error if not registered 'cantResolve(type:, name:)'
}

func init() throws {
  let person = try container.throwsResolve(Person.self)
}
```

You can guarantee not optional result or use ```try!``` and ```try?``` if needs

jakubvano:
What are the usecase, in which it is beneficial to use `throwsResolve` instead of current `resolve`?

You could rewrite your examples with current API IMO without  loosing clarity.
```swift
guard let person = container.resolve(Person.self) else { 
    // do error handling
}
person.play()

func init() throws {
    guard let person = container.resolve(Person.self) else { throw SomeError() }
}
```
There has been some discussion about implementing throwable `resolve` in the past (see #99, [#6](https://github.com/Swinject/SwinjectAutoregistration/issues/6)), and at that time we came to the conclusion that changing the API is not wort it.

I wouldn't mind reopening this discussion - personally I would like throwing `resolve()` methods a bit more as well. 
However I don't think it is a good idea to have 2 kinds of API - if we wanted to have `throwsResolve`, it should probably replace current `resolve` API.

Lumyk:
Think that ```throws``` better than ```Optional``` and more evidently in use. Hope you can find better solution to make all ```throws``` based. Ready to help if needs ;)

PR #349, opened 2018-05-16
Container as a Protocol to fascilitate mocking

doozMen:
Swift is a protocol based language. In one project I faced the problem that it was difficult to write tests with a real container. I wanted to mock it out but because it was not a protocol this was difficult. I had to make some functions and types public as a downside of this approach. What do you think?


jakubvano:
Closing due to inactivity

PR #350, opened 2018-05-26
Use NSRecursiveLock instead of NSLock

gilroykilroy:
If you have a re-entrant resolve (i.e. when resolving "A" using synchronized(), "A" tries to resolve "B") it will deadlock:

"You should not use this class to implement a recursive lock. Calling the lock method twice on the same thread will lock up your thread permanently. Use the NSRecursiveLock class to implement recursive locks instead."

### DIFF
--- a/Sources/Info.plist
+++ b/Sources/Info.plist
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>2.4.0</string>
+	<string>2.4.1</string>
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleVersion</key>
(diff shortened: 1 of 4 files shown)


serges147:
Finally! I did almost the same in PR #269 exactly 1 year ago - it was rejected back then.

serges147:
In my PR #269 I had corresponding unit test as well. @gilroykilroy would you like to try to reincarnate it as well?

gilroykilroy:
Feel free.  I was totally unaware of this previous fix.  I just came across the problem in my project and needed it fixed so I fixed it.

jakubvano:
I have added tests for this scenarion in #353.

@serges147 Sorry for taking this long to address this issue. Back then it looked like we might be able to implement more robust thread safety model, which would more comprehensively address this problem.
However, we weren't able to do that yet, and quite a few people are hindered by it - that is why including this workaround now seems like a better option.

Thanks for your work guys!

Issue #368, opened 2018-08-29
Optional types on Xcode 10 (Swift 4.2)


Axadiw:
Hey, I've got a piece of code that was working perfectly fine in XCode 9 (Swift 4.0), but on XCode 10 (Swift 4.2) it resolves to `nil` only.

I'm registering my own type in the assembly with:

```
container.register(Optional<MyType>.self) {
            (_, myParam: string) in
...

return instance
        }
```

and resolving it like

```
let myInstance = resolver.resolve(Optional< MyType >.self,
                        arguments: "abc")
```

As I've mentioned above, on Xcode this code resolves correctly (to instance of `MyType`) but on Xcode 10 i'm receiving just `nil`. 

Intresting fact is that after placing a breakpoint inside the register clojure, debugger doesn't stop there

taher-mosbah:
@Axadiw see #367 , I think we are having the same issue, please let us know if you identified the cause of this bug.

Axadiw:
Because I don't have `Swinject: Resolution failed. Expected registration` error on the console - I thought that this might be similar, but different issue.

Anyways - until it would be fixed I've used quick workaround - I've just wrapped my object with a struct like

```
struct MyTypeBox {
let myType: MyType?
}
```

and I'm registering `MyTypeBox` in swinject, without an optional

romainbiard:
same issue on my side.
I can't really use this workaround since I'm using an injected  `weak var`(to avoid retain cycle) and thus need to be optional :-/

Issue #367, opened 2018-08-20
Resolution failed in Xcode 10


taher-mosbah:
Xcode version 10.0 beta 6 (10L232m)
Swinject version: 2.4.0

I'm registering some of my dependencies like this
```
            container.register(ConnectedCoordinatorModel?.self) { _ in
                if let user = UserManager.currentUser() {
                    return ConnectedCoordinatorModelImp(user)
                }
                return nil
            }
```

And then resolving like this

```
    public static func resolve<Service>(_ serviceType: Service.Type) -> Service? {
        let container = DependenciesContainer.shared.container
        if let value = container.resolve(serviceType) {
            return value
        }
        if let value = container.resolve(Optional<Service>.self) {
            return value
        }
        return nil
    }
```
this code used to work perfectly but in Xcode 10 I'm having the following log:
```
Swinject: Resolution failed. Expected registration:
	{ Service: ConnectedCoordinatorModel, Factory: Resolver -> ConnectedCoordinatorModel }
Available registrations:
```

po container.description gives the following 
```
{ { Service: Optional<ConnectedCoordinatorModel>, Factory: Resolver -> Optional<ConnectedCoordinatorModel>, ObjectScope: graph }}
```

PR #371, opened 2018-09-17
Fix tests in XCode 10

janhalousek:
Fixes failing tests in XCode 10

This MR is superset of https://github.com/Swinject/Swinject/pull/369 and probably resolves issues #368, #367 (haven't tested all cases).

The problem lies in casting nil variable to _Any_ type, which seems to be a new thing in XCode 10 / Swift 4.2
```
class Bar {}
var foo: Bar? = nil

if let fooUnwrapped = foo as? Any {
    // this passes and fooUnwrapped is Optional(nil)
}
```

### DIFF
--- a/Sources/InstanceStorage.swift
+++ b/Sources/InstanceStorage.swift
@@ -120,7 +120,10 @@ private class Weak<Wrapped> {
     }
 #else
     var value: Wrapped? {
-        get { return object as? Wrapped }
+        get {
+            guard let object = object else { return nil }
+            return object as? Wrapped
+        }
         set { object = newValue as AnyObject? }
     }
 #endif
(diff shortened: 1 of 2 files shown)


Sources/InstanceStorage.swift
```diff
@@ -120,7 +120,10 @@ private class Weak<Wrapped> {
     }
 #else
     var value: Wrapped? {
-        get { return object as? Wrapped }
+        get {
+            guard let object = object else { return nil }
+            return object as? Wrapped
+        }
```
jakubvano: This fix also needs to be applied to the linux branch

janhalousek: Adjusted in ce3b62e702d8fb59a41999353c8a257878adc035

jakubvano: 👍

jakubvano:
Thanks for the PR! 👍 
I will try to release Swift 4.2 compatible version ASAP

PR #379, opened 2018-11-12
Documentation improvements #trivial

gobetti:
Hi! I had a great time reading your docs, I just thought that a few minor changes would have made my understanding slightly easier, so why not to share it. I've tried to isolate the changes as much as possible in separate commits, so just let me know if there's anything you'd like me to revert. Thanks!

### DIFF
--- a/Documentation/CircularDependencies.md
+++ b/Documentation/CircularDependencies.md
@@ -72,7 +72,7 @@ container.register(ChildProtocol.self) { _ in Child() }
     }
 ```
 
-Here both or either of the depending properties must be specified in the `initCompleted` callback to avoid infinite recursion.
+Here both or either of the depending properties must be specified in the `initCompleted` callback to avoid infinite recursion, however using it on both [has proven to avoid issues](#remark).
 
 ## Initializer/Initializer Dependencies
 
--- a/Documentation/InjectionPatterns.md
+++ b/Documentation/InjectionPatterns.md
@@ -44,7 +44,7 @@ Note that the actual type of `Animal` passed to the initializer of `PetOwner` is
 
 ## Property Injection
 
-_Property injection_ is a pattern to pass a dependency to a dependent instance via a setter property. Property injection is appropriate if the dependency is optional to the dependent instance.
+_Property injection_ is a pattern to pass a dependency to a dependent instance via a setter property. Property injection is appropriate if the dependency is optional to the dependent instance, and is required in the case of _[circular dependencies](CircularDependencies.md)_.
 
 The following code defines property injection to `PetOwner2`:
 
--- a/Documentation/InjectionPatterns.md
+++ b/Documentation/InjectionPatterns.md
@@ -40,7 +40,7 @@ class PetOwner: Person {
 }
 ```
 
-Note that the actual type of `Animal` passed to the initializer of `PetOwner` is automatically resolved by Swinject when the `PetOwner` instance is created:
+Note that the actual type of `Animal` passed to the initializer of `PetOwner` is automatically resolved by Swinject when the `PetOwner` instance is created.
 
 ## Property Injection
 
--- a/Documentation/InjectionPatterns.md
+++ b/Documentation/InjectionPatterns.md
@@ -82,7 +82,7 @@ container.register(Person.self) { r in
 
 ## Method Injection
 
-_Method injection_ is a similar pattern to _property injection_, but it uses a method to pass dependencies to a dependent instance.
+_Method injection_ is a similar pattern to _property injection_, but it uses a method to pass dependencies to a dependent instance. It is appropriate when the dependency is encapsulated as a private property.
 
 The following code defines Method Injection to `PetOwner3`:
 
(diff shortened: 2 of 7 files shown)


Documentation/CircularDependencies.md
```diff
@@ -72,7 +72,7 @@ container.register(ChildProtocol.self) { _ in Child() }
     }
 ```
 
-Here both or either of the depending properties must be specified in the `initCompleted` callback to avoid infinite recursion.
+Here both or either of the depending properties must be specified in the `initCompleted` callback to avoid infinite recursion, however using it on both [has proven to avoid issues](#remark).
```
jakubvano: I wouldn't highlight the issues in this section. Put like this it suggests that using `initCompleted` is preferable, which IMO should not be the best practice - we should prefer constructor injection as much as possible. Moreover, in my experience, those issues are usually negligible.

Documentation/InjectionPatterns.md
```diff
@@ -40,22 +40,22 @@ class PetOwner: Person {
...
-container.register(Person.self) { r in
-    let owner = PetOwner2()
-    owner.pet = r.resolve(Animal.self)
-    return owner
-}
+container.register(Person.self) { _ in PetOwner2() }
+    .initCompleted { r, p in
+        let owner = p as! PetOwner2
+        owner.pet = r.resolve(Animal.self)
+    }
```
jakubvano: Original registration example should be prefered, as it is simple and does not require casting.

Documentation/InjectionPatterns.md
```diff
@@ -68,39 +68,39 @@ class PetOwner2: Person {
 }
 ```
 
-Or, you can use <a name="initialization-callback">`initCompleted` callback</a> instead of defining the injection in the registration closure:
+Note that the <a name="initialization-callback">`initCompleted` callback</a> in some cases is the same as defining the injection in the registration closure (see the following example), but `initCompleted` is preferred since it is able to resolve _[circular dependencies](CircularDependencies.md)_.
```
jakubvano: It is true that it is necessary for circular dependencies, but i wouldn't say that it is therefore prefered - it often requires type casting, which looses type safety

Documentation/InjectionPatterns.md
```diff
@@ -68,39 +68,39 @@ class PetOwner2: Person {
...
-container.register(Person.self) { r in
-    let owner = PetOwner3()
-    owner.setPet(r.resolve(Animal.self)!)
-    return owner
-}
+container.register(Person.self) { _ in PetOwner3() }
+    .initCompleted { r, p in
+        let owner = p as! PetOwner3
+        owner.setPet(r.resolve(Animal.self)!)
+    }
```
jakubvano: Original registration method should be preffered

Documentation/CircularDependencies.md
gobetti: this is inside the section "Property injection", from my understanding it is preferred in this scenario, isn't it? This is exactly what I missed the most while reading the docs, once you reach the circular dependencies doc, it is clear that `initCompleted` avoids issues, but until you get there, the docs make you think that it is syntax sugar for setting the property in the registration closure

Documentation/InjectionPatterns.md
gobetti: however once you have circular dependencies, you'll have to refactor it as mentioned in the circular dependencies doc

Documentation/InjectionPatterns.md
gobetti: I know this is out of the scope of this PR, but is it possible to use generics to have the correct type as a parameter in the initCompleted closure?

Documentation/CircularDependencies.md
jakubvano: You are right, until reader get's to circular dependencies, `initCompleted` does not have a clear purpose. It has other utilities (it heavily relied on e.g. in `SwinjectStoryboard`), but circular deps is its main purpose...
Maybe we could fix that by adding reference to circular dependencies when showcasing this syntax in previous chapters?

Documentation/InjectionPatterns.md
jakubvano: Not necessarily - you would need to refactor only one definition (in the case of 2 related objects).

Documentation/InjectionPatterns.md
jakubvano: It is quite problematic - mostly because you cannot require relationships between generic parameters: Something like
```
func foo<T, U>(type: T.Type) -> U where U: T
```
is not possible

Documentation/InjectionPatterns.md
gobetti: right, the docs mention you should change to `initCompleted` "in either or both". That was the keypoint that made me think I should open a PR (if always using it has no problems, and never using it may run into problems, then I would choose to always use it).

But if it's not the preferred way, and users should remember to switch to `initCompleted` only once they have problems, then I think this PR doesn't make much sense. Sorry about that!

PR #391, opened 2019-02-09
Swift 5/Swift 4.2 Compatibility Fix

tcldr:
Enables Swinject to be run successfully as a dependency in a Swift 5 (Xcode 10.2b2) project.

Changes made:
- Migrate hashing functions to use Swift 4.2s `hash(into:)`.
- Remove redundant access modifier to eliminate Swift 5 warnings
- Bump Quick/Nimble private deps to latest release versions

All tests run successfully under Xcode 10.1 with the release versions of Quick and Nimble.

However, tests will not run using Xcode 10.2b2 as the current release versions of Quick and Nimble do not support the Swift 5 compiler.

This problem won't be encountered by people using Swinject as a dependency because Quick/Nimble are located in the private Cartfile.

Nevertheless, a quick test via temporarily bumping Quick/Nimble to their master branches (which support Swift 5 compiler) shows all tests complete successfully on all testable targets.

### DIFF
--- a/Cartfile.private
+++ b/Cartfile.private
@@ -1,2 +1,2 @@
-github "Quick/Quick" ~> 1.1
-github "Quick/Nimble" ~> 7.0
+github "Quick/Quick" ~> 2.0
+github "Quick/Nimble" ~> 8.0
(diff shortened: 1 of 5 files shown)


jakubvano:
👍 Thanks for the PR! Looks good.
Travis now has Xcode 10.2 beta environment available, so we can bump the config. And once the Quick / Nimble have the compatible official releases ready we can merge this.

tcldr:
Pleasure! Sounds good.

tcldr:
This should be good go now. Latest versions of Quick/Nimble.

jakubvano:
👍 indeed

PR #403, opened 2019-05-15
Correct mistyped doc comments

lucasxyang:
Correct mistyped doc comments at Resolver.swift


jakubvano:
Thanks for the improvement 👍

PR #409, opened 2019-05-24
Properly reset graph object scope on parent container

jakubvano:


florentdouine:
Thank you very much :) 
Capture d’écran 2019-06-04 à 15 48 01

PR #414, opened 2019-06-06
fix-swift-package-manager-integration

thbonk:
With Swift 5.0 it wasn't possible to define Swinject as a dependency. Fixed Package.swift


jakubvano:
Thanks for the PR!

PR #421, opened 2019-07-04
[Experimental] Property wrappers

jakubvano:
Draft of the possible API extension taking advantage of new `@propertyWrapper` functionality.

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,2 +1,2 @@
-github "Quick/Nimble" "v8.0.0"
-github "Quick/Quick" "v2.0.0"
+github "Quick/Nimble" "v8.0.2"
+github "Quick/Quick" "v2.1.0"
--- a/Sources/Container.swift
+++ b/Sources/Container.swift
@@ -236,8 +236,10 @@ extension Container: _Resolver {
 
     fileprivate func incrementResolutionDepth() {
         parent?.incrementResolutionDepth()
+        implicitResolver = self
         if resolutionDepth == 0 && currentObjectGraph == nil {
             currentObjectGraph = GraphIdentifier()
+            implicitResolver = self
         }
         guard resolutionDepth < maxResolutionDepth else {
             fatalError("Infinite recursive call for circular dependency has been detected. " +
@@ -256,7 +258,8 @@ extension Container: _Resolver {
 
     fileprivate func graphResolutionCompleted() {
         services.values.forEach { $0.storage.graphResolutionCompleted() }
-        self.currentObjectGraph = nil
+        currentObjectGraph = nil
+        implicitResolver = nil
     }
 }
 
--- a/Sources/Injected.swift
+++ b/Sources/Injected.swift
@@ -24,9 +24,28 @@ public struct Injected<T> {
         wrappedValue = implicitResolver!.resolve(T.self, argument: argument)!
     }
 
-    public init<A1, A2>(arguments a1: A1, _ a2: A2) {
-        wrappedValue = implicitResolver!.resolve(T.self, arguments: a1, a2)!
+    public init<A1, A2>(arguments arg1: A1, _ arg2: A2) {
+        wrappedValue = implicitResolver!.resolve(T.self, arguments: arg1, arg2)!
     }
 
     // TODO: init overloads
 }
+
+@propertyWrapper
+public struct LazyInjected<T> {
+    @Injected private var lazy: Lazy<T>
+
+    public var wrappedValue: T { return lazy.instance }
+
+    public init() {}
+}
+
+@propertyWrapper
+public struct ProviderInjected<T> {
+    @Injected private var lazy: Provider<T>
+
+    public var wrappedValue: T { return lazy.instance }
+
+    public init() {}
+}
+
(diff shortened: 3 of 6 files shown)


CharlieSuP1:
Great idea!   
And also I thought I would be good to add keyPath param to Inject constructor.
Like
 `
public struct Injected<T> {
    public init<Root, Value>(_ keyPath: KeyPath<Root, Value>) {
    }
}
`
The usage would be like
` @Injected private var @Injected(\AccountService.userName)`
Thus minimum dependency can be achieved.

jakubvano:
@CharlieSuP1 I'm not sure I follow - how would we utilize the `keyPath`?

CharlieSuP1:
> @CharlieSuP1 I'm not sure I follow - how would we utilize the `keyPath`?

I am actually using @KeyPath in my project.

```
@propertyWrapper
public struct KeyPath<Value>  {
    public let wrappedValue: Value

    public init<Root>(_ keyPath: Swift.KeyPath<Root, Value>) {
        self.init(Injected<Root>(), keyPath)
    }

    public init<Root>(_ injected: Injected<Root>, _ keyPath: Swift.KeyPath<Root, Value>) {
        self.wrappedValue = injected.wrappedValue[keyPath: keyPath]
    }
}
```

The basic usage would be:

```
class DemoVC: UIViewController {

    @KeyPath(\AccountService.userName)
    private var name: String

}
```

This will first resolve AccountService(with no arguments) from container and get the value of userName property of AccountService.

If the resolution from container need argument, then caller need to manually pass an Injected object.

The usage would be:
```
    @KeyPath(Injected(arguments: "arg1", "arg2"), \AccountService.userName)
    private var anotherName: String
```

The purpose of the @KeyPath is for minimum dependency.
For example, In my project, there are many places need to get bool value of one particular feature gating key.
Instead of injecting whole FeatureGatingService (which contains lots of feature gating key), I would like to use @KeyPath(\FeatureGatingService.searchEnable) var searchEnable: Bool to inject only one bool value.

jakubvano:
Thanks for the explanation, I think I understand your usecase.

Wouldn't it then be cleaner not to declare a dependency on the `AccountService` at all? Something like
```
// magic strings could be abstracted away
register(String.self, tag: "userName") { $0.resolve(AccountService.self).userName }

class DemoVC {
    @Injected(tag: "userName") var name: String
}
```

CharlieSuP1:
Yes, register String type with a tag can be a solution.

And In my project the there are lots of cases that some class need to depend on the AccountService, other classes need to depend only on or two properties of AccountService.

So with you approach I have to register both AccountService and the properties of AccountService.

And the management of the tag is not easy and can be a problem.

One way I come up with is to make a rule like :
anyone who register a property of a class( or a struct ), you should register a tag with that property type, and the tag name should be ModuleName + ClassName + PropertyName.

So the callers know how to retrieve that property.

But this way has some drawbacks:

1. This involves lots of registration.
2. From the callers perspective, the caller doesn't know which properties are registered.

So if I am the author of AccountService module, I have to make a statement. like 
"Ok, everyone listen, I have registered the username and userid property of AccountService in the Container. So anyone who want that, you can get with the tag XXXX.  "

This can be error prone.

jakubvano:
I get your points - even though it is more of a service locator than DI feature, we should be able to do this as a purely additive feature without any impact on the core implementation.

There already is a rudimentary implementation for this in 3.0 branch  [here](https://github.com/Swinject/Swinject/blob/3.0/Sources/3.0%20API/Retrieval.swift), and [here](https://github.com/Swinject/Swinject/blob/3.0/Tests/3.0%20API/RetrievalSpec.swift).

From my point of view key path resolution would fit into this as a special kind of `InstanceRequest` (it could be generalized & turned into protocol), with construction analogous to current [instance request syntax](https://github.com/Swinject/Swinject/blob/3.0/Sources/3.0%20API/InstanceRequest%2BAPi.swift), something like
```
protocol ResolvableRequest {
   associatedtype Result
   func fulfill(with resolver: Resolver) throws -> Result
}

extension InstanceRequest: ResolvableRequest {}

struct KeyPathRequest<Root, Value>: ResolvableRequest {}

func keyPath<Root, Value>(_ KeyPath<Root, Value>, ..) -> KeyPathRequest <Root, Value> {}

extension Injected {
   init<Request>(_ request: Request) where Request: ResolvableRequest {}
}
```

which would be then used as

```
class DemoVC {
    @Injected(keyPath(\AccountService.userName)) 
    var userName: String
}
```
possibly with some convenience initalizers on `Injected`

CharlieSuP1:
Thanks for reply!
> possibly with some convenience initalizers on Injected
Yes,  I actually added the keypath param to Injected constructor.
I will look through the 3.0 branch of Swinject to find out additions.

One thing that I wand to ask is that there are @Injected in both branch 3.0 and in this PR.
What are their connections?  
I mean is this PR will eventually merged into master or branch 3.0 will be merged into master?

jakubvano:
>I mean is this PR will eventually merged into master or branch 3.0 will be merged into master?

This PR was only experimental, 3.0 implementation is more relevant and not connected

PR #422, opened 2019-07-06
Configuration via JSON

Goxtend:
Hey there,

I added functionality to use a JSON-File as the configuration.
its not perfect though, swift doesn't support reflection that well.. But it does its job :)

I would appreciate your feedback
Greetings from Germany


Goxtend:
Travis CI says there are some loose files in the package but i cant figure out why.. has someone an idea?

Also the Coding-Convention bot says the name ContainerSpec_Config is invalid because of non alphanumeric characters. But the other test-classes also have the '_'-Character in their names?

Sorry for the inconvenience.. its my first creating an pull request..

jakubvano:
Thanks for the PR 👍 
What is the motivation behind this feature? I haven't used DI framework with external config in any language yet - what are the advantages of the DI configuration in JSON over code in general?

My initial thoughts are that this is not a good fit for Swinject, mainly because: 
- as you mentioned, Swift does not have required required level of reflection, so this would be restricted to the code build on ObjC
- this loses type safety and compile time catching of bugs

Maybe this could be an optional extension (perhaps integrated with [SwinjectPropertyLoader](https://github.com/Swinject/SwinjectPropertyLoader)?

yoichitgy:
Thanks @Goxtend for proposing the nice feature👍 and sorry for taking to long time to respond🙇‍♂️

As @jakubvano said, it's nice to have this kind of additional features, yet I agree it's nice, in another repo like SwinjectPropertyLoader, SwinjectAutoregistration or Swinject-CodeGen. We made Swinject extensible when we made SwinjectPropertyLoader exported from Swinject repo. I think JSON config can be also implemented in a similar way ;)

Let me close this PR🙏 Thank you again for your motivation to use Swinject😃

PR #426, opened 2019-07-22
Update documentation

gfpoliva:
In Container Hierarchy documentation, the next page is called **Properties** and leading to **Modularizing Service Registration**.

Don't you guys think that it should be more understandable to be called `Modularizing Service Registration (Assembly)`, like we have in Table of Contents?

If so, this Pull Request does this small improvement in the documentation.


jakubvano:
Thanks for the improvement 👍

Issue #432, opened 2019-08-30
Bundle id Error


raptorxcz:
**Describe the bug**
Report from Apple:
ITMS-90806: CFBundleIdentifier collision - Each bundle must have a unique bundle identifier. The bundle identifier 'com.el-eleven.Swinject' is used in the bundles '[Swinject.framework, Swinject.framework]'
App have iOS and watchOS target.

**To Reproduce**
1. make app with multiple platforms
2. upload it to AppStore

**Expected behavior**
No error from Apple

**Environment:**
 - Xcode version: Xcode 10.3
 - Swift version: 5.0.1
 - OS running SWXMLHash: MacOS 10.14.6 running Xcode

**Additional context**
Add `.${PLATFORM_NAME}` to end of bundle id

jakubvano:
Resolved by #433, released in 2.7.0

PR #433, opened 2019-09-02
Separate bundle id for platforms

raptorxcz:
Solves issue https://github.com/Swinject/Swinject/issues/432.


jakubvano:
Thanks for the PR! 👍

PR #442, opened 2019-12-18
A quick Swift5 property wrapper

Tyler-Keith-Thompson:
I scanned through current issues and noticed that somebody had played with this idea before and that it *might* be in 3.0. That being said I'd really love this functionality to be present and I believe we could get a quick win.

Different considerations:
If it'd be reasonable to ship a shared container with the library this property wrapper could be a little nicer:
```swift
@propertyWrapper
public struct DependencyInjected<Value> {
    let name:String?
    public init(wrappedValue value: Value?) {
        self.name = nil
    }
    public init(wrappedValue value: Value?, name:String?) {
        self.name = name
    }

    public lazy var wrappedValue: Value? = {
        Container.shared.resolve(Value.self, name: name)
    }()
}
```

This means that when using this in a class you don't need to add "= nil"
Example:
```swift
class SomeClass {
    @DependencyInjected var someType:SomeProtocol?
}
```

Instead of:
```swift
class SomeClass {
    @DependencyInjected var someType:SomeProtocol? = nil
}
```

As an aside one of my biggest issues with dependency injection frameworks is the "magic". I love that this one is relatively easy to follow and heavily makes use of swift generics. I believe utilizing a property wrapper like this can help get rid of some of the "magic" introduced with SwinjectStoryboard.

You can command + click on this property wrapper and follow through the implementation to understand what container is being used to resolve a dependency. 

NOTE: I am a big fan of this framework but have only started playing with it recently. It's possible if not likely that I've missed important functionality here.

### DIFF
--- a/Sources/DependencyInjectionPropertyWrapper.swift
+++ b/Sources/DependencyInjectionPropertyWrapper.swift
@@ -7,10 +7,18 @@
 
 import Foundation
 @propertyWrapper
+/// Wrapper to enable lazy resolution of a type from a container.
+/// Supply the container and an optional name
 public struct DependencyInjected<Value> {
-    let name:String?
-    let container:Container
+    ///Optional name to pass to container for resolution
+    private let name:String?
+    ///The container to resolve from
+    private let container:Container
     
+    /// init(wrappedValue: container: name): Wrapper to enable lazy resolution of a type from a container
+    /// - Parameter wrappedValue: This should always be set to nil, if you do pass a value it will be ignored in favor of what is in the container
+    /// - Parameter container: A `Container` the property should be resolved from
+    /// - Parameter name: An optional name to pass for more specific resolution from the container.
     public init(wrappedValue value: Value? = nil, container:Container, name:String? = nil) {
         self.name = name
         self.container = container
--- a/Sources/DependencyInjectionPropertyWrapper.swift
+++ b/Sources/DependencyInjectionPropertyWrapper.swift
@@ -1,0 +1,22 @@
+//
+//  DependencyInjectionPropertyWrapper.swift
+//  Swinject
+//
+//  Created by Tyler Thompson on 12/17/19.
+//
+
+import Foundation
+@propertyWrapper
+public struct DependencyInjected<Value> {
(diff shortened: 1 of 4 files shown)


Sources/DependencyInjectionPropertyWrapper.swift
```diff
@@ -0,0 +1,22 @@
...
+//  Created by Tyler Thompson on 12/17/19.
+//
+
+import Foundation
+@propertyWrapper
+public struct DependencyInjected<Value> {
+    let name:String?
+    let container:Container
+    
+    public init(wrappedValue value: Value? = nil, container:Container, name:String? = nil) {
```
davidthorn, line 14: Please add documentation for each of the parameters. I understand that they are already pretty logical, but its better to have documentation for all public methods and their parameters.

Sources/DependencyInjectionPropertyWrapper.swift
```diff
@@ -0,0 +1,22 @@
...
+//  DependencyInjectionPropertyWrapper.swift
+//  Swinject
+//
+//  Created by Tyler Thompson on 12/17/19.
+//
+
+import Foundation
+@propertyWrapper
+public struct DependencyInjected<Value> {
+    let name:String?
```
davidthorn, line 11: Please declare the properties private and add documentation for each of them. Thanks

davidthorn (changes requested):
Please correct the changes that I have requested first.

tkohout:
I think this is a great feature to have, we should think about the API naming well because we will have to keep it for the future versions.

This is what @jakubvano has now in the WIP 3.0 branch now:

https://github.com/Swinject/Swinject/blob/bde873b53251a57edc069aaf098b22f18f203de4/Sources/2.0%20API/InstanceWrapper.swift

Tyler-Keith-Thompson:
Alright, I've added the requested documentation and also brought it up to date with master, given it's been a while. I'm not quite clear on the API naming conventions to use. `Provider` (which is used in the linked blob) seems to be a different concept, if I'm understanding correctly.

PR #445, opened 2020-02-25
Add synchronizedResolver property to Assembler

### LINKED ISSUE #221: Assembly/Assembler need thread safety

devxoul:
- Resolves #221

### DIFF
--- a/Documentation/Assembler.md
+++ b/Documentation/Assembler.md
@@ -132,6 +132,12 @@ Now you can resolve any components from either assembly:
 let fooManager = assembler.resolver.resolve(FooManagerProtocol.self)!
 ```
 
+For [thread safety](ThreadSafety.md), you may use `synchronizedResolver`:
+
+```swift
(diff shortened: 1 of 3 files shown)


opfeffer:
Whats the status of this PR? Would be very happy to see this land? cc/ @yoichitgy

bryanwagner:
This would have been handy to have at the moment for me as well; I could create an extension function if only `container` wasn't private.

ursusursus:
why did this die?

peterkos:
Any updates? cc @mpdifran

maxim-chipeev:
Hey @peterkos, with #514 and #523 this should be resolved on 2.8.3.

PR #451, opened 2020-04-04
Enables support for distributing Swinject as a binary library

### LINKED ISSUE #450: Missing support for binary frameworks with module stability
`Swinject.xcodeproj` isn't configured to generate a `.swiftinterface` file which means dynamic frameworks can only be used with projects using the same version of the Swift compiler.

It would be nice to not have to recompile the framework when a new Swift compiler is released.

devioustree:
This fixes #450.

**Solution**

To allow frameworks to be used by multiple Swift version, we need to set:

```
BUILD_LIBRARY_FOR_DISTRIBUTION = YES
```

This isn't necessary for a Swift package (since they are build from source), but it's useful for Carthage or other means of distributing binary frameworks.

This was added to the `project.yml` and then `xcodegen` was run.

**Testing**

I built `Swinject.framework` in Xcode 11.3.1. Then in Xcode 11.4 I created a test project and linked to the generated project. It built and linked successfully.

**Tools**

```zsh
$ xcodegen --version
Version: 2.14.0

$ swift --version
Apple Swift version 5.1.3 (swiftlang-1100.0.282.1 clang-1100.0.33.15)
Target: x86_64-apple-darwin19.3.0

$ carthage version
0.34.0
```


yoichitgy:
Thank you for making the PR😃 and sorry for taking long time to merge it🙇‍♂️ I was  not aware about `BUILD_LIBRARY_FOR_DISTRIBUTION` setting. I appreciate your PR❤️

yoichitgy:
I actually took only the changes about `BUILD_LIBRARY_FOR_DISTRIBUTION` to avoid conflict.
https://github.com/Swinject/Swinject/commit/190a6d1e4c7c60258034d9b3b8098d87a2daa8fd

PR #465, opened 2020-10-28
Add dynamic option for the library to SPM

mpdifran:
This is a pattern I've seen in a few other libraries to get around Xcode's lack of support for choosing a dynamic library when linking to a project.


yoichitgy (approved):
LGTM😃👍 Thank you for adding the option🚀

PR #466, opened 2020-10-30
Add SynchronizedResolver property to Resolver protocol

itsthejb:
- Alternate resolution to #221 / #445 
- Expose a _`SynchronizedResolver`_ property in the `Resolver` protocol.
- We use the Viper pattern, and have the architecture that a `Resolver` might be passed into some components (e.g. the Module), so that other components can be dynamically generated:

```
    public func assemble(container: Container) {
        container.register(FooModule.self) { resolver in
            FooModuleImplementation(resolver: resolver.synchronizedResolver)
        }
    }
```

- This causes thread-safety problems, however
- This PR is a _possible_ solution, although it is a little untidy:
  - Even better would be somehow adding configurability/refactoring, [so that the `SynchronizedResolver` also resolves dependencies using itself](https://github.com/Swinject/Swinject/blob/master/Sources/SynchronizedResolver.swift#L20-L36)
  - Currently a `SynchronizedResolver` only works for the initial `resolve(...)` call

Happy to discuss other options, but we will be using this as a fix for the time being

### DIFF
--- a/Sources/Container.swift
+++ b/Sources/Container.swift
@@ -261,6 +261,9 @@ extension Container: _Resolver {
 // MARK: - Resolver
 
 extension Container: Resolver {
+
+    public var synchronizedResolver: Resolver { synchronize() }
+
     /// Retrieves the instance with the specified service type.
     ///
     /// - Parameter serviceType: The service type to resolve.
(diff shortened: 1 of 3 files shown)


talron10:
Did it work for you? It seems like this will create a few synchronized resolvers while there should be only one, we are considering using this solution too

itsthejb:
@talron10 Yes, it works - not crashing. But it's _far_ from ideal, more like a workaround for now. If you have serious stability issues, it might be worth doing in the short-term, though

PR #467, opened 2020-12-18
Feature/type safe naming

laviallb:
Add type safety to registration/resolving. 
When registering elements identified by name, we might leave literal strings in the code and/or keep a list of string identifying registered element. But there is not safety in case of changing an element type at registration without updating resolving, or any other way of switching types/string.

This PR adds:
* `TypeIdentifier` protocol to associate a type to a string name
* `Identifiable` protocol to statically bind a registrable type to a`TypeIdentifier`
* `DedicatedIdentifier` to bind types which can not extend a protocol such as protocols
* entensions over `Container` and `Resolver` to use this type safe identification on top of literal string naming

### DIFF
--- a/Sources/Container.Arguments.erb
+++ b/Sources/Container.Arguments.erb
@@ -1,9 +1,5 @@
 //
-//  Container.Arguments.swift
-//  Swinject
-//
-//  Created by Yoichi Tagaya on 8/18/15.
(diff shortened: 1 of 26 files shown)


laviallb:
Hello @jakubvano, what do you think about this PR?

felginep:
Any update on this ? @yoichitgy maybe ?

PR #470, opened 2021-04-17
SECURITY.md

asepabdaz:


yoichitgy:
Thanks @asepabdaz for making the PR and letting us be aware of a security policy.

For now, let us close this PR because we don't have our security policy. Before making a `SECURITY.md` file, we should make our policy.

Issue #472, opened 2021-07-01
Remove dependency on Quick/Nimble for better maintainability and easier collaborator onboarding


yoichitgy:
Remove dependency on Quick/Nimble and use XCTest for the following reasons.

- Get dev environment setup easier for collaborator onboarding
- Get rid of updates of Quick/Nimble to catch up with the latest Xcode/Swift versions
- Reduce CI time

Although it makes conflicts on PR #431 for Swinject 3.0, we take the priority of maintainability because we couldn't make releases for a few years.

Quick/Nimble are lovable frameworks for developers❤️, but it's good to take the reasons above because our maintainers' capacity is limited😣

yoichitgy:
Merged  #473 🚀

Issue #309, opened 2017-12-07
Please deactivate codeCoverageEnabled to prevent appstore rejection


BennX:
Please deactivate codeCoverageEnabled inside of the Swinject-iOS.xcscheme else a carthage dependency build will result in an rejection of Apple for the reason:

> Invalid Bundle - Disallowed LLVM instrumentation. Do not submit apps with LLVM profiling instrumentation or coverage collection enabled. Turn off LLVM profiling or code coverage, rebuild your app and resubmit the app.

current possible workaround:
https://stackoverflow.com/questions/46160518/xcode-9-carthage-itunes-connect-error-invalid-bundle-disallowed-llvm-instr

s4cha:
Issued https://github.com/Swinject/Swinject/pull/311 to solve this issue :)

jakubvano:
It has turned out to be the Carthage issue, and has been fixed in [Carthage 0.26](https://github.com/Carthage/Carthage/releases/tag/0.26.0).

Issue #296, opened 2017-09-28
Undefined symbols for architecture x86_64: ___llvm_profile_runtime


andriko13:
I'm getting an issue linking Swinject framework into my project when using Xcode9 and Swift3.2

It's giving me the following error:
```
Undefined symbols for architecture x86_64:
  "___llvm_profile_runtime", referenced from:
      ___llvm_profile_runtime_user in Swinject(SpinLock.o)
      ___llvm_profile_runtime_user in Swinject(Assembler.o)
      ___llvm_profile_runtime_user in Swinject(ObjectScope.Standard.o)
      ___llvm_profile_runtime_user in Swinject(ServiceEntry.o)
      ___llvm_profile_runtime_user in Swinject(SynchronizedResolver.o)
      ___llvm_profile_runtime_user in Swinject(DebugHelper.o)
      ___llvm_profile_runtime_user in Swinject(InstanceStorage.o)
      ...
     (maybe you meant: ___llvm_profile_runtime_user)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Getting this error with both 2.1.1 and 2.0.0 stable releases
<img width="753" alt="screen shot 2017-09-28 at 9 03 56 am" src="https://user-images.githubusercontent.com/15827184/30970891-fdb294a2-a42b-11e7-9a31-5ce3f6bc349a.png">

evisoup:
Try cleaning up  the project. Also, release for `Xcode9 and Swift3.2` is not out yet.

andriko13:
Tried cleaning project, derived data, etc. to no avail.
I understand the release isn't out yet; just wanted to make sure this issue is tracked.

yoichitgy:
@andriko13, how did you install Swinject? With Carthage or CocoaPods?

Timmmm:
This is related to profiling. I get the exact same error in a different project when compiling with `-fprofile-instr-generate`. No idea why yet - I guess Clang is not linking with some require runtime library but I can't work out which one.

Timmmm:
Ah, found the problem. If you pass `-fprofile-instr-generate` during compilation, but not during linking, Clang won't link with the appropriate libraries and you get this error.

jakubvano:
@andriko13 is your original issue related to what @Timmmm is describing? If that is the case, and I understand it correctly, this is not the issue with the Swinject, but rather with specific project settings?

Reflejo:
[...]

PR #473, opened 2021-07-01
Remove dependency on Quick/Nimble

yoichitgy:
This PR addresses issue #472.

- Replaced `XxxSpec` using Quick/Nimble with `XxxTests` using XCTest.
- Removed dependency on Quick/Nimble from Cartthage and SPM configs, and resulted in removal of dependency on Carthage.
- (This is not directly related to the issue) but fixed Xcode project setting to exclude SwiftLint.yml from Unit Test packages because those were included unnecessarily.


72A12F4E:
Hopefully this fixes my issues. I haven't been able to build locally & test on an iOS simulator because of  some issue with Quick & Nimble not setting up correctly when I `carthage bootstrap`

72A12F4E (changes requested):
One recommendation I would make is to enable code coverage for each scheme. When I checked out this PR locally, code coverage was disabled. 

I enabled it locally, and found that we're at about 72%.

yoichitgy:
Thanks @72A12F4E for your comment! It's a nice idea to keep the code coverage visible to be aware to write unit tests👍 

It looked Swinject users saw some issues (#296 #309). Maybe it's nice to turn on code coverage measurement on CI or somewhere for contributors/maintainers💪

mpdifran (approved):

mpdifran:
Tested resolving this branch via Carthage and SPM and everything works!

yoichitgy:
Thank you for your kind and careful review😃 I merged the PR🚀

PR #484, opened 2021-07-07
Renamed SpinLock to RecursiveLock

### LINKED ISSUE #459: Is SpinLock actually a spin lock?
[`SpinLock`](https://github.com/Swinject/Swinject/blob/master/Sources/SpinLock.swift) seems like a misleading name because under the hood, it's using `NSRecursiveLock` which, as far as I can tell, is a blocking lock.

Mike Ash, writing about a list that includes `NSRecursiveLock`, [states](https://www.mikeash.com/pyblog/friday-qa-2017-10-27-locks-thread-safety-and-swift-2017-edition.html):

> I mentioned spinlocks as one type of lock, but none of the APIs listed here are spinlocks.

Apple's synchronization guide [states](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/ThreadSafety/ThreadSafety.html) about spin locks:

> The system does not provide any implementations of spin locks because of their polling nature, but you can easily implement them in specific situations.

1ucas:
This PR resolves #459 

- SpinLock and its file were renamed to RecursiveLock.

Since it's an internal property there aren't any breaking changes.

### DIFF
--- a/Sources/Container.swift
+++ b/Sources/Container.swift
@@ -26,7 +26,7 @@ public final class Container {
     private let debugHelper: DebugHelper
     private let defaultObjectScope: ObjectScope
     internal var currentObjectGraph: GraphIdentifier?
-    internal let lock: SpinLock // Used by SynchronizedResolver.
+    internal let lock: RecursiveLock // Used by SynchronizedResolver.
     internal var behaviors = [Behavior]()
 
     internal init(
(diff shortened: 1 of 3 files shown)


yoichitgy (approved):
Thank you for making the PR😃 LGTM👍👍👍

Issue #486, opened 2021-07-08
Fix release makefile to get the current branch in a way that supports older git versions


mpdifran:

1ucas:
I still have the same doubt. Shouldn't it (the makefile) be a workflow at the CI, in a way it doesn't depend on each developer's environment?

mpdifran:
If it's easy enough to do it through CI, then we can jump straight there. I was figuring this would be an easy first step that would help make releases easier, and we can integrate into CI as a phase 2.

1ucas:
That a look at this Pull Request. It has almost all step.
But I didn't know it would be worth it to invest more time.
https://github.com/1ucas/Swinject/pull/3/files

yoichitgy:
That's nice! Thanks 1ucas!

mpdifran:
@yoichitgy I believe you've made some updates in this area, should this task be closed?

yoichitgy:
@mpdifran I noticed fixing our Makefile was insufficient and made the following PRs.

- https://github.com/Swinject/Swinject/pull/495
- https://github.com/Swinject/SwinjectStoryboard/pull/167

Would you review them? After merging them, we can close this issue👍

I think the release process using CI/CD (GitHub Actions) can be another (next step) issue.

mpdifran:
Reviewed and merged! I'll close this issue now.

PR #489, opened 2021-07-16
Update makefile to support older versions of git

mpdifran:
Relevant task: Swinject/Swinject#486


yoichitgy (approved):
LGTM! Thanks for fixing it😃

PR #503, opened 2021-09-14
Update Container to inherit NSObject

kimxwan0319:
For versatility, I updated `Container` to inherit from `NSObject`.

Actually, I want use [Then](https://github.com/devxoul/Then). (very awesome sugar)
Like this.
```swift
let container = Container().then {
   $0.register(Animal.self) { _ in Cat(name: "Mimi") }
}
```
Isn't it really that simple?

If we inherit from NSObject, we can do cool things like this.
Thanks.

### DIFF
--- a/Sources/Container.swift
+++ b/Sources/Container.swift
@@ -19,7 +19,7 @@ import Foundation
 ///
 /// where `A` and `X` are protocols, `B` is a type conforming `A`, and `Y` is a type conforming `X`
 /// and depending on `A`.
-public final class Container {
+public final class Container: NSObject {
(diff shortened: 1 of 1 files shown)


yoichitgy:
Thanks @kimxwan0319 for suggesting the nice syntax sugar!

Although your suggestion is really nice, but I would like to keep Swinject as pure Swift. As far as I read the README of [Then](https://github.com/devxoul/Then), you can achieve the same functionality with an extension.

> // Want to use with your own types? Just make extensions.
> 
> extension MyType: Then {}
> 
> let instance = MyType().then {
>   $0.really = "awesome!"
> }

PR #514, opened 2022-07-29
Internalize `Container` synchronization

welshm:
- Add synchronization tests for `Lazy` and `Producer`
- Fix issue of thread safety not enforced for `Lazy` and `Producer`
- Removes need for `SynchronizedResolver` by internalizing usage of the `RecursiveLock`
  - Should be more performant as the synchronization is just done around the final resolution instead of all entry points

Should fix issue https://github.com/Swinject/Swinject/issues/493

### DIFF
--- a/Sources/Container.swift
+++ b/Sources/Container.swift
@@ -25,6 +25,7 @@ public final class Container {
     private var resolutionDepth = 0
     private let debugHelper: DebugHelper
     private let defaultObjectScope: ObjectScope
+    private let synchronized: Bool
     internal var currentObjectGraph: GraphIdentifier?
     internal let lock: RecursiveLock // Used by SynchronizedResolver.
     internal var behaviors = [Behavior]()
@@ -32,12 +33,14 @@ public final class Container {
     internal init(
         parent: Container? = nil,
         debugHelper: DebugHelper,
-        defaultObjectScope: ObjectScope = .graph
+        defaultObjectScope: ObjectScope = .graph,
+        synchronized: Bool = false
     ) {
         self.parent = parent
         self.debugHelper = debugHelper
         lock = parent.map { $0.lock } ?? RecursiveLock()
         self.defaultObjectScope = defaultObjectScope
+        self.synchronized = synchronized
     }
 
     /// Instantiates a ``Container``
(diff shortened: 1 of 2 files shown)


welshm:
@yoichitgy I would appreciate a review if you have time!

Sources/Container.swift
```diff
@@ -293,32 +308,42 @@ extension Container: Resolver {
...
-
-        if let persistedInstance = persistedInstance(Service.self, from: entry, in: currentObjectGraph) {
-            return persistedInstance
-        }
-
-        let resolvedInstance = invoker(entry.factory as! Factory)
-        if let persistedInstance = persistedInstance(Service.self, from: entry, in: currentObjectGraph) {
-            // An instance for the key might be added by the factory invocation.
-            return persistedInstance
+        let resolution: () -> Service? = { [weak self] in
```
maxim-chipeev, line 313: Maybe strong [self] is okay to capture here? Reference will be release when we exit scope and that way can omit all the `self.*` for readability.

welshm: I have no objections to that - makes sense since it's within the same scope. Will wait for further comments before I make that change

mpdifran (approved):
Looks good to me! Would we also need to add `let synchronized: Bool` to `Assembler` as well? Or would we do something like this:
```
myAssembler.resolver.synchronize()
```

welshm:
> Looks good to me! Would we also need to add `let synchronized: Bool` to `Assembler` as well? Or would we do something like this:
> 
> ```
> myAssembler.resolver.synchronize()
> ```

I don't think it's needed on the Assembler because the documentation suggests not to synchronize until you've finished registering

```
/// Returns a synchronized view of the container for thread safety.
/// The returned container is ``Resolver`` type. Call this method after you finish all service registrations
/// to the original container.
///
/// - Returns: A synchronized container as ``Resolver``.
```

So I think what you've suggested with `myAssembler.resolver.synchronize()` would be the correct approach - unless there is a goal to synchronize assembly as well.

Sources/Container.swift
welshm: Actually on looking at this again, I think it should stay `weak`.

In the event that the thread is locked waiting for the lock to be acquired and `self` is deallocated, this should exit without returning a value IMO.

maxim-chipeev: Is it possible that the compiler will allow for `Container` to be deinitialized before `resolve<Service, Factory>` exits? I ask because all class-level functions are passed `self` as an argument at a lower level.

welshm: Hmm... you might be right. Because we're locking on the execution of the function, the reference to self is likely preserved either way. Updated

maxim-chipeev (approved):

Sources/Container.swift
maxim-chipeev: Ok—could you remove the `self.` references from within the block for style? It's fine if you keep them just figured they're redundant now

PR #517, opened 2022-08-30
Add test target to SPM package

bradfol:
Allows running unit tests when loading up the project from the package definition

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -20,5 +20,7 @@ let package = Package(
     targets: [
         .target(name: "Swinject",
                 path: "Sources"),
+        .testTarget(name: "SwinjectTests",
+                    dependencies: ["Swinject"]),
     ]
 )


Package.swift
```diff
@@ -20,5 +20,8 @@ let package = Package(
     targets: [
         .target(name: "Swinject",
                 path: "Sources"),
+        .testTarget(name: "SwinjectTests",
+                    dependencies: ["Swinject"],
+                    path: "Tests/SwinjectTests"),
```
oronbz, line 25:
```suggestion
```
Seems redundant since it's already on the default path.

oronbz: Regardless, a blessed PR

bradfol: Ah thanks for pointing that out, updated the PR to remove the custom path

maxim-chipeev (approved):
@mpdifran This looks good too.

PR #521, opened 2022-09-18
Add  `hasRegistration` method

oronbz:
## Motivation
Sometimes you just want to check if `Service` has been registered before you `register` it yourself:
```Swift
class OrderAssembly: Assembly {
        container.autoregister(OrderViewModel.self, initializer: OrderViewModel.init)
        
        // Register mock if no registration was already made
        if container.resolve(SplitFarePresenting.self) == nil {
            container.autoregister(SplitFarePresenting.self, initializer: SplitFarePresenterMock.init)
        }
    }
}
```

`resolve` could trigger other unwanted side effects or resource allocation which you may not want to trigger.

Hence, `hasRegistration` method trying to solve  it:
```Swift
if container.hasRegistration(SplitFarePresenting.self) == nil {
            container.autoregister(SplitFarePresenting.self, initializer: SplitFarePresenterMock.init)
        }
```


oronbz:
HI, any chance that this can be merged? Is Swinject still maintained?

Sources/Container.swift
```diff
@@ -176,6 +176,23 @@ public final class Container {
...
+    ///                  that have the same service and factory types.
+    ///
+    /// - Returns: A ``Bool`` which represents whether or not the ``Service`` has been registered.
+    public func hasRegistration<Service>(
+        _ serviceType: Service.Type,
+        name: String? = nil
+    ) -> Bool {
+        getRegistrations().contains { key, _ in
+            key.serviceType == serviceType && key.name == name
+        }
```
maxim-chipeev, line 194: Just thinking about this possible scenario:
```
// What kind of registration?
//
// Could it be helpful to know if it has a 0-arg registration 
// vs an n-arg registration & which types?
if container.hasRegistration(TypeWithArgs.self) {
  container.register( … )
}
```
Maybe you could use the ServiceKey struct here and compare it against `services[key]` instead? Should be a bit more performant too for larger projects.

maxim-chipeev (commented):
If you register an identical ServiceKey it will override the existing registration (if there is one). But, I do think this PR does offer at least one benefit here, which is perhaps sometimes you want to provide a 'default' Assembly in a modular project, only if there hasn't been one registered already in that Container.

Sources/Container.swift
oronbz: Hi, exactly because of the reason you mentioned here: 
> If you register an identical ServiceKey it will override the existing registration (if there is one)

This is crucial (at least for me). We have lazily loaded assemblies that lives on other modules, and are loaded some time after the main container of the app has registered everything. On this scenario I don't want them to override the current registrations.

I've checked the `ServiceKey` way before implementing it the way I did, but assembling one from scratch just to compare is a tough one.

oronbz: Regarding 0-arg vs n-arg, I don't think it is relevant for the scenario, because those variations are relevant to the implementations of the service and not for it's protocol.

maxim-chipeev: Thanks for the quick reply. In that case, perhaps naming the function `hasAnyRegistration` will be more clear.

For your specific scenario I think this is sufficient. But, for the purpose of lazy Assemblies that shouldn't overwrite a _specific_ registration—the natural extension of this PR would be to add `hasRegistration` which could check against a ServiceKey. It could also be quite useful, if you're interested. :)

oronbz: @maxim-chipeev  `ServiceKey` is an internal type, and not exposed to `Swinject`'s consumers (:

oronbz: In any case, I've updated the API as you suggested to: `hasAnyRegistrations(of serviceType)`

oronbz: In any case, I've updated the API as you suggested to: hasAnyRegistrations(of serviceType)

maxim-chipeev: right! 😅 I should've put it a different way—to check against the Service type, name, & argument types, this doesn't require exposing the ServiceKey

maxim-chipeev (approved):
Thanks for making the changes, this will be a good addition

Sources/Container.swift
oronbz: Yeah, so I'm already checking against type and name, but argument types is pretty unintuitive to assemble, so we might need a different set of generic variations, per each argument count, similar to autoRegister.

mpdifran (approved):

PR #523, opened 2022-10-15
GraphIdentifier threadsafety optimization & patch interleaving race condition in Lazy resolve

maxim-chipeev:
We encountered a bizarre error in an async context; this PR offers a solution to two problems. (A topic of note here is that the `GraphIdentifier` is a _class_. We will come back to this later).

- GraphIdentifier dereference crash, e.g. `EXC_BAD_ACCESS: KERN_INVALID_ADDRESS: Attempted to dereference garbage pointer`
- Race condition wherein an async call to `Lazy.instance` could acquire the lock, set the graph identifier, release the lock, then **re-acquire** the lock and begin resolving. In between the lock release/acquisition of one `.instance` call, another `.instance` call could update the GraphIdentifier before the first thread could call `resolve` and re-acquire the lock. Can results in unexpected behaviour.

## Synchronization
Synchronizing a container will emit a child which ensures that any async calls to `resolve(…)` should happen sequentially and never concurrently (or more precisely, for the logic within).

One line here is quite interesting:
CleanShot 2022-10-15 at 13 07 49@2x

It happens within the sync context, so it should be safe. However, there is one Swinject object that _also_ contains a graph identifier reference. That is Lazy.

CleanShot 2022-10-15 at 12 37 11@2x

CleanShot 2022-10-15 at 12 39 05@2x

A Lazy property can be held by any class. Once that class is dereferenced/deallocated so are its children, in this case the `ObjectIdentifier` being one of them.

Critically, **we have no control over when a Lazy property (and its GraphIdentifier ref) might be released**.

```
deinit {
    container?.lock.sync {
        // THIS AVOIDS THE DEREFERENCE CRASH
        self.graphIdentifier = nil
    }
}
```
- To further my point, the above line entirely prevents the crash when added to the `Lazy` type.

## Solution

This one fairly simple and results meets expectations for GraphIdentifier behaviour. It should be an Identifiable struct, the performance tradeoffs here are negligible: a smidge costlier to create, but faster to pass by value (requires no ARC calls).

### Some debug artifacts…
- Swift ARC concurrency crashes are really hard to debug…so in this case it helps to have an extra pair of eyes and some compiler help
- [neat Swift forum thread on `@_assemblyVision`](https://forums.swift.org/t/psa-compiler-optimisation-remarks/54152)

CleanShot 2022-10-15 at 12 26 06@2x

CleanShot 2022-10-15 at 12 42 57@2x
CleanShot 2022-10-15 at 12 44 26@2x

### DIFF
--- a/Sources/Container.swift
+++ b/Sources/Container.swift
@@ -225,8 +225,8 @@ extension Container: _Resolver {
         )
 
         if let entry = getEntry(for: key) {
-            let factory = { [weak self] (graphIdentifier: GraphIdentifier?) in
-                let action = { [weak self] in
+            let factory = { [weak self] (graphIdentifier: GraphIdentifier?) -> Any? in
+                let action = { [weak self] () -> Any? in
                     if let graphIdentifier = graphIdentifier {
                         self?.restoreObjectGraph(graphIdentifier)
                     }
--- a/Sources/Container.swift
+++ b/Sources/Container.swift
@@ -177,6 +177,8 @@ public final class Container {
         behaviors.append(behavior)
     }
 
+    /// Restores the object graph to match the given identifier.
+    /// Not synchronized, use lock to edit safely.
     internal func restoreObjectGraph(_ identifier: GraphIdentifier) {
         currentObjectGraph = identifier
     }
--- a/Sources/Container.swift
+++ b/Sources/Container.swift
@@ -178,16 +178,7 @@ public final class Container {
     }
 
     internal func restoreObjectGraph(_ identifier: GraphIdentifier) {
-        let action = { [weak self] in
-            self?.currentObjectGraph = identifier
-        }
-        if synchronized {
-            lock.sync {
-                action()
-            }
-        } else {
-            action()
-        }
+        currentObjectGraph = identifier
     }
 }
 
@@ -234,7 +225,19 @@ extension Container: _Resolver {
         )
 
         if let entry = getEntry(for: key) {
-            let factory = { [weak self] in self?.resolve(entry: entry, invoker: invoker) as Any? }
+            let factory = { [weak self] (graphIdentifier: GraphIdentifier?) in
+                let action = { [weak self] in
+                    if let graphIdentifier = graphIdentifier {
+                        self?.restoreObjectGraph(graphIdentifier)
+                    }
+                    return self?.resolve(entry: entry, invoker: invoker) as Any?
+                }
+                if self?.synchronized ?? true {
+                    return self?.lock.sync(action: action)
+                } else {
+                    return action()
+                }
+            }
             return wrapper.init(inContainer: self, withInstanceFactory: factory) as? Wrapper
         } else {
             return wrapper.init(inContainer: self, withInstanceFactory: nil) as? Wrapper
--- a/Sources/Container.swift
+++ b/Sources/Container.swift
@@ -227,9 +227,9 @@ extension Container: _Resolver {
         if let entry = getEntry(for: key) {
             let factory = { [weak self] (graphIdentifier: GraphIdentifier?) in
                 let action = { [weak self] in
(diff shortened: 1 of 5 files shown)


maxim-chipeev:
@mpdifran @welshm Seems I can't add reviewers, but I think you might like this. 🙂

Sources/GraphIdentifier.swift
```diff
@@ -2,6 +2,14 @@
 //  Copyright © 2019 Swinject Contributors. All rights reserved.
 //
 
+#if os(iOS) || os(macOS)
```
mpdifran, line 5: Why only on these platforms?

Sources/InstanceStorage.swift
```diff
@@ -35,7 +35,7 @@ public final class GraphStorage: InstanceStorage {
         self.instance = instance
 
         if instances[graph] == nil { instances[graph] = Weak() }
-        instances[graph]?.value = instance
+        instances[graph]?.value = instance as AnyObject?
```
mpdifran, line 38: Won't this result in nil when using the struct?

Sources/InstanceStorage.swift
```diff
@@ -106,25 +106,6 @@ public final class CompositeStorage: InstanceStorage {
...
-    #else
-        var value: Wrapped? {
-            get {
-                guard let object = object else { return nil }
-                return object as? Wrapped
-            }
-            set { object = newValue as AnyObject? }
-        }
-    #endif
+private class Weak {
```
mpdifran, line 109: Why remove the generic type?

maxim-chipeev: It wasn't used.

Sources/InstanceStorage.swift
maxim-chipeev: Replicates the old behavior—references to structs were discarded.

Sources/GraphIdentifier.swift
maxim-chipeev: Linux didn't recognize `UUID()`. I see a library for `UniqueID()` available but it requires adding a dependency. We could go that route I think.

maxim-chipeev: https://forums.swift.org/t/uniqueid-time-ordered-uuids-in-swift/53265

Sources/Container.swift
```diff
@@ -178,16 +178,7 @@ public final class Container {
     }
 
     internal func restoreObjectGraph(_ identifier: GraphIdentifier) {
```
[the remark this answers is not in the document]
maxim-chipeev: True, and if it weren't for the tests we could mark this private too. Or discard the function and just set it—since this function is only called within this file at this point. I will add a comment.

Sources/GraphIdentifier.swift
mpdifran: Does that mean Linux will still have threading issues?

mpdifran: Might be good to call that out at the top of the PR. Do we also want to fix this on `tvOS` and `watchOS`?

maxim-chipeev: True! Will do that

Actually, I think can we just leverage random `Int.max` for the Linux case & have that fixed here too?

maxim-chipeev:
what the, why did clicking the request button delete the other ones… 🤯

Sources/InstanceStorage.swift
maxim-chipeev: reverted to avoid any issues since this PR should still work without these changes

Sources/GraphIdentifier.swift
mpdifran: That's probably fine.

mpdifran (approved):

PR #537, opened 2023-08-16
Remove generated header file to make it a pure swift library 

sanju-naik:
Bazel natively doesn't provide a rule to build mixed language (Swift + ObjC ) framework hence removing the .h inclusion from pod spec and also deleting the header file which Xcode creates when you create a framework target.

The Library was anyway all swift so deleting this header shouldn't create an issue. 

This MR also bumps the iOS deployment target to iOS 11 to resolve podspec update issue with XCode 14.3


sanju-naik:
@maxim-chipeev Can I please get a review on this? Thanks.

sanju-naik:
@maxim-chipeev Gentle bump on this 🙏

lgreydev:
Nice

PR #540, opened 2023-10-09
Make synchronous reads and writes to the container safe

### LINKED ISSUE #539: Synchronous reads and writes fail
The documentation makes it clear that `Container` is not thread safe, but reads can be made thread safe by calling `.synchronize()` to get a `Resolver`. That is true if they are on their own. However, if you are writing to the container, the reads in the `Resolver`s are not necessarily thread safe.

This test demonstrates the issue:

```swift
    func testSynchronizedResolverHighCounts() {
        let container = Container()
        
        DispatchQueue.global(qos: .background).async {
            for index in 0...100_000 {
                container.register(Animal.self, factory: { _ in
                    Cat(name: "\(index)")
                })
            }
        }
        
        DispatchQueue.global(qos: .background).async {
            DispatchQueue.concurrentPerform(iterations: 100_000) { (index) in
                let cat = container.synchronize().resolve(Animal.self)
                print(String(describing: cat))
            }
        }
    }
```

Usually the test does pass. When it fails it's normally in the function `graphResolutionCompleted`, but I don't believe it's specific to that and could fail in other places.

dalemyers:
The `services` parameter is not thread safe. Concurrent reads are safe, but concurrent reads with writes happening at the same time are not. This adds a read/write lock around it, making reads synchronous and safe, but writes block to ensure safety.

Resolves #539

### DIFF
--- a/Sources/Container.swift
+++ b/Sources/Container.swift
@@ -20,7 +20,7 @@ import Foundation
 /// where `A` and `X` are protocols, `B` is a type conforming `A`, and `Y` is a type conforming `X`
 /// and depending on `A`.
 public final class Container {
-    internal var services = [ServiceKey: ServiceEntryProtocol]()
+    internal var services = ThreadSafeDictionary<ServiceKey, ServiceEntryProtocol>()
     private let parent: Container? // Used by HierarchyObjectScope
     private var resolutionDepth = 0
     private let debugHelper: DebugHelper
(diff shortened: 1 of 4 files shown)


Goles:
Seeing the same thing. Thanks for working on this.

dalemyers:
@mpdifran / @maxim-chipeev Any chance of a review on this? Thanks! 😄

dalemyers:
@mpdifran / @maxim-chipeev Just bumping this again as we are still seeing it and we'd really rather not have to fork.

maxim-chipeev (approved):
Thanks for the follow-up ping—and for adding the functionality, I will merge this in for the use-case of concurrent writes to the container.

Issue #542, opened 2023-12-12
Privacy Manifest


FONickReichard:
At WWDC23 Apple announced that apps and SDKs that make use of certain "required reason" APIs will need to provide a privacy manifest.

Is this update on the roadmap for the team? I appreciate that enforcement won't happen until Spring 2024 but I wanted to make contact so that we can plan our own app releases to take this into account.

Apple News:

https://developer.apple.com/news/?id=r1henawx 

Thank you

kishorepran:
Here’s some useful references:

https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_data_use_in_privacy_manifests

https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api

https://developer.apple.com/videos/play/wwdc2023/10060/

kishorepran:
Here’s some useful references:

https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_data_use_in_privacy_manifests

https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api

https://developer.apple.com/videos/play/wwdc2023/10060/
CC: @mpdifran

mpdifran:
cc'ing @maxim-chipeev as well!

zarghol:
Hello, as the deadline for compliance is fixed as the May 1st, is there any updates on this subject ?
Thanks !

iharkatkavets:
Hey, any updates guys?

maxim-chipeev:
Hi, definitely will be taking a look very soon at this!

kishorepran:
Check with this tool as well. Swinject is not using any required reason API.
https://github.com/omarzl/ios_17_required_reason_api_scanner/tree/binary_analyzer
@FONickReichard  : you can close this issue safely

PR #550, opened 2024-03-26
Adds privacy manifest

maxim-chipeev:
The privacy manifest discussion doesn't apply to Swinject since this library does not engage in any of the relevant activities. Swinject does not have a way neither ever intends to use data on disk / or any kind of data linked to a user.

Our manifest can just specify that Swinject doesn't handle user data or make the special listed API calls.
- #549 
- #542 

More info here.
- [Privacy Manifest Files](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files)


yoichitgy:
(LGTM🙌)

PR #551, opened 2024-04-04
Undo graph restoration once Lazy wrapper resolves its local hierarchy

maxim-chipeev:
Fixes this possible edge case.
1. Container begins to resolve complex object.
2. Complex object contains multiple graph scoped items, lazily wrapped.
3. A lazy instance is accessed early, before the graph resolution has completed.
4. Lazy wrapper restores the stored graph identifier.
5. Container continues resolution using the incorrect graph identifier for _this_ resolution.


Andre113 (approved):
Nice! Thanks for solving!

PR #567, opened 2024-05-29
Implement ContainerCapture configuration

bradfol:


Sources/Container.swift
```diff
@@ -31,17 +31,27 @@ public final class Container {
...
+        case owningContainer
+    }
+
     internal init(
         parent: Container? = nil,
         debugHelper: DebugHelper,
         defaultObjectScope: ObjectScope = .graph,
-        synchronized: Bool = false
+        synchronized: Bool = false,
+        containerCapture: ContainerCapture
```
skorulis-ap, line 47: Should have a default here, probably .legacy in the short term.

Sources/Container.swift
```diff
@@ -58,9 +71,10 @@ public final class Container {
         parent: Container? = nil,
         defaultObjectScope: ObjectScope = .graph,
         behaviors: [Behavior] = [],
-        registeringClosure: (Container) -> Void = { _ in }
+        registeringClosure: (Container) -> Void = { _ in },
+        resolverHierarchyAccess: ResolverHierarchyAccess = .receiver
```
bradfol, line 75: I left the default here as the current behavior, but it would be nice to discuss changing the default in a future version.

Tests/SwinjectTests/ContainerTests.swift
```diff
@@ -90,6 +90,69 @@ class ContainerTests: XCTestCase {
...
+        child.register(Animal.self, factory: { _ in Cat()})
+
+        parent.register(Animal.self, name: "Spot", factory: { resolver in
+            resolver.resolve(Animal.self)!
+        })
+        .inObjectScope(.container)
+
+        XCTAssert(child.resolve(Animal.self, name: "Spot") is Cat)
+        // The instance provided by the child leaks into the parent
+        XCTAssert(parent.resolve(Animal.self, name: "Spot") is Cat)
```
bradfol, line 122: Also want to call attention to this test case as this behavior is the most unexpected/problematic of the way things currently work

skorulis-ap (approved):

maxim-chipeev:
Will look at this soon to see if I can merge this as part of a new release cut.

bradfol:
When testing this change @skorulis-ap discovered that it does not work properly with the way that synchronized containers are set up. This is due to the part where the synchronized container contains no service entries, all entries are on the unsynchronized container, which means the modified `getEntry()` method will find the unsynchronized container when it does its look up, which is problematic.

PR #568, opened 2024-05-31
Exposes object graph

maxim-chipeev:
At Faire we use a custom property wrapper `@Resolved` to make Swinject calls (hopefully we can land it in this repo soon), and had to make some significant performance trade-offs to cache the GraphIdentifier without it being properly exposed (involved a workaround using the Lazy wrapper).

This one method & a simple ServiceKey check won't cause any change in Swinject native performance but will permit a few powerful use cases. :)

### DIFF
--- a/Sources/Container.swift
+++ b/Sources/Container.swift
@@ -39,7 +39,7 @@ public final class Container {
     ) {
         self.parent = parent
         self.debugHelper = debugHelper
-        lock = parent.map { $0.lock } ?? RecursiveLock()
+        lock = parent.map(\.lock) ?? RecursiveLock()
         self.defaultObjectScope = defaultObjectScope
         self.synchronized = synchronized
     }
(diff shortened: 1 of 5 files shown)


Sources/Container.swift
```diff
@@ -224,6 +238,10 @@ extension Container: _Resolver {
             var resolvedInstance: Service?
             let key = ServiceKey(serviceType: Service.self, argumentsType: Arguments.self, name: name, option: option)
 
+            if key == Self.graphIdentifierKey {
+                return currentObjectGraph as! Service?
```
zjradfor, line 242: We will crash if this cast fails.
Would it make sense to throw an error with a message instead? Or default to a nil Service?

zjradfor (approved):

Sources/Container.swift
maxim-chipeev: Hm, when would this happen? The cast is always to `GraphIdentifier?` as `graphIdentifierKey` ensures we're making a call to resolve a GraphIdentifier.

If the `currentObjectGraph` is value `nil` that will successfully cast to `GraphIdentifier?`.

lucasmarcal-faire, line 242:
```suggestion
                return currentObjectGraph as? Service
```
Has similar runtime performance

PR #570, opened 2024-07-15
Add Container registration functions for @MainActor factories

skorulis-ap:
Adds an API to register services marked as `@MainActor` without build warnings or errors when strict concurrency is enabled. 

Examples of how this can be used to register a `@MainActor` isolated service.
```swift
container.register(MainService.self, mainActorFactory: { _ in MainService() })
container.register(MainService.self, factory: { _ in MainService() })
```

The resolution of services does not change but will crash if attempted off the main thread, effectively giving runtime concurrency safety. These new APIs will not be picked up automatically so there is no risk of a crash without a developer opting in to use strict concurrency.

The top 2 registrations demonstrate the new behaviour when using strict concurrency while the bottom demonstrates the old.

Screenshot 2024-07-15 at 4 22 06 PM

### DIFF
--- a/Sources/Container.MainActor.erb
+++ b/Sources/Container.MainActor.erb
@@ -1,0 +1,82 @@
+//
+//  Copyright © 2024 Swinject Contributors. All rights reserved.
+//
+
+//
+// NOTICE:
+//
+// Container.MainActor.swift is generated from Container.MainActor.erb by ERB.
+// Do NOT modify Container.MainActor.swift directly.
+// Instead, modify Container.MainActor.erb and run `script/gencode` at the project root directory to generate the code.
+//
+
+<% arg_count = 9 %>
+import Foundation
+
+// MARK: - MainActor registration
+@available(iOS 13.0, macOS 10.15, *)
+extension Container {
+
+    /// Adds a registration for the specified service with the factory closure to specify how the service is
+    /// resolved with dependencies which must be resolved on the main actor.
+    ///
+    /// - Parameters:
+    ///   - serviceType:        The service type to register.
+    ///   - name:               A registration name, which is used to differentiate from other registrations
+    ///                         that have the same service and factory types.
+    ///   - mainActorFactory:   The @MainActor closure to specify how the service type is resolved with the dependencies of the type.
+    ///                         It is invoked when the ``Container`` needs to instantiate the instance.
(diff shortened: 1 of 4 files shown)


Sources/Container.MainActor.swift
skorulis-ap, line 1: All code in this file is generated from the erb

script/gencode
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-files="Sources/Container.Arguments Sources/SynchronizedResolver.Arguments Sources/Resolver Sources/ServiceEntry.TypeForwarding"
```
skorulis-ap, line 3: This wasn't functioning and hasn't been run in a while, I only committed the changes for the files I modified.

Sources/Container.MainActor.erb
```diff
@@ -0,0 +1,82 @@
...
+    @discardableResult
+    public func register<Service>(
+        _ serviceType: Service.Type,
+        name: String? = nil,
+        mainActorFactory: @escaping @MainActor (Resolver) -> Service
+    ) -> ServiceEntry<Service> {
+        return register(serviceType, name: name) { r in
+            MainActor.assumeIsolated {
+                return mainActorFactory(r)
+            }
```
g-mark, line 42: Only calling `MainActor.assumeIsolated` will crash when called _off_ of the main thread.

To avoid a crash, you could check to see if execution is already on the main thread, and then hop to it if needed:
```suggestion
            onMainActor {
               mainActorFactory(r)
            }
```

Where onMainActor is defined as:
```swift
func onMainActor<ReturnValue>(work: () -> ReturnValue) -> ReturnValue {
    if Thread.isMainThread {
        return MainActor.assumeIsolated { work() }
    }
    return DispatchQueue.main.sync {
        MainActor.assumeIsolated { work() }
    }
}
```

maxim-chipeev:
Hi @skorulis-ap, I may have sometime this week to double-check this change—is there a chance you could write tests to validate it is still fine to register types off main-thread as per @g-mark's comment? The [documentation](https://developer.apple.com/documentation/swift/mainactor/assumeisolated(_:file:line:)) suggests this may be impossible.

As for the solution offered, I hesitate to dispatch to main using `sync`…that might open a whole can of deadlock worms.

Issue #547, opened 2024-03-05
Synchronized containers are prone to deadlock when reading from multiple threads


dalemyers:
If you try and read a container from multiple threads, even if you've called `.synchronize()`, you will run into a dead lock issue. The following code is an example:

```swift

let container = Container()

_ = container.register(Foo.self) { _ in
    Thread.sleep(forTimeInterval: 100.0)
    return Foo()
}.inObjectScope(.container)

_ = container.register(Bar.self) { _ in
    return Bar()
}.inObjectScope(.container)

DispatchQueue.global().async {
    _ = container.synchronize().resolve(Foo.self)
}

// Allow the dispatch to happen
Thread.sleep(forTimeInterval: 0.5)

// This will block for 100 seconds
_ = container.synchronize().resolve(Bar.self)
```

The problem stems from the use of `NSRecursiveLock`. This lock allows multiple read acquisitions on the same thread, but will block across different threads.

Fab1n:
This is not really a deadlock, this is just a lock, but not locking indefinitely.

PR #574, opened 2024-11-05
Fix ThreadSafety when registering and resolving at the same time on different threads.

Sali0m:
Hello, as mentioned in this issue: https://github.com/Swinject/Swinject/issues/547
We encountered on our project some crashes when we try to resolve some dependencies while we are registering others on different threads. 

Here is what the stack trace of the different threads looks like (second one is the one crashing in this case).
image
image

Using the previously used ThreadSafeDictionary seem to fix the issue.

### DIFF
--- a/Sources/Container.swift
+++ b/Sources/Container.swift
@@ -20,7 +20,7 @@ import Foundation
 /// where `A` and `X` are protocols, `B` is a type conforming `A`, and `Y` is a type conforming `X`
 /// and depending on `A`.
 public final class Container {
-    internal var services = [ServiceKey: ServiceEntryProtocol]()
+    internal var services = ThreadSafeDictionary<ServiceKey, ServiceEntryProtocol>()
     private let parent: Container? // Used by HierarchyObjectScope
     private var resolutionDepth = 0
     private let debugHelper: DebugHelper
@@ -297,7 +297,7 @@ extension Container: _Resolver {
 
     fileprivate func getRegistrations() -> [ServiceKey: ServiceEntryProtocol] {
         var registrations = parent?.getRegistrations() ?? [:]
-        services.forEach { key, value in registrations[key] = value }
+        services.forEachRead { key, value in registrations[key] = value }
         return registrations
     }
 
--- a/Sources/ThreadSafeDictionary.swift
+++ b/Sources/ThreadSafeDictionary.swift
@@ -8,6 +8,12 @@ internal final class ThreadSafeDictionary<KeyType: Hashable, ValueType> {
 
     private let lock = ReadWriteLock()
 
+    public var values: Dictionary<KeyType, ValueType>.Values {
+        lock.read {
+            self.internalDictionary.values
+        }
+    }
+
     // MARK: - Initializers
 
     public init(dictionary: [KeyType: ValueType]? = nil) {
@@ -31,7 +37,7 @@ internal final class ThreadSafeDictionary<KeyType: Hashable, ValueType> {
             }
         }
     }
-    
+
     public subscript(key: KeyType, default defaultValue: @autoclosure () -> ValueType) -> ValueType? {
         get {
             return lock.read { self.internalDictionary[key, default: defaultValue()] }
@@ -49,10 +55,16 @@ internal final class ThreadSafeDictionary<KeyType: Hashable, ValueType> {
             self.internalDictionary.forEach(block)
         }
     }
-    
+
     public func forEachWrite(_ block: ((key: KeyType, value: ValueType)) -> Void) {
         lock.write {
             self.internalDictionary.forEach(block)
+        }
+    }
+
+    public func contains(where predicate: ((key: KeyType, value: ValueType)) -> Bool) -> Bool {
+        lock.read {
+            self.internalDictionary.contains(where: predicate)
         }
     }
 


Sali0m:
> Hi guys! @Sali0m, would you believe that is missing just the lint and spm-check to be able to merge this Pull Request? I'm asking it because my application is needing this fix too. I'd be happy to help merging this PR.

I have no idea why the checks haven't run tbh, if you have any idea how I can make them run, please do tell :)

Sali0m:
> Hi guys! @Sali0m, would you believe that is missing just the lint and spm-check to be able to merge this Pull Request? I'm asking it because my application is needing this fix too. I'd be happy to help merging this PR.

I need an approval from a maintainer for the actions to run

marcellyluise:
> > Hi guys! @Sali0m, would you believe that is missing just the lint and spm-check to be able to merge this Pull Request? I'm asking it because my application is needing this fix too. I'd be happy to help merging this PR.
> 
> I need an approval from a maintainer for the actions to run

Oh I see. Let's try adding then to this thread: @yoichitgy, @jakubvano, @1ucas and @maxim-chipeev you guys are the top 4 commuters here. Any chance any of you are also a maintainer to take a look in this PR? We need a maintainer approval so the actions run to validate this pull request.

maxim-chipeev:
Hi, could you share a bit more on the nature of the problem? ThreadSafeDictionary is not performant enough for the needs of very-large codebases.

Is there a chance you can reproduce the issue with tests?

nathan-batista:
Hello @maxim-chipeev.
I'm facing the same issue and here is a code snippet of a test that is able to replicate the issue.
I don't know if their use case is the same as mine but the cause would be the same, an access to the services dictionary from different threads.

`func test_SwinjectCrash() {
        let numberOfQueues = 25
        var expectations = [XCTestExpectation]()
        
        let resolutionDate = Date() + 1
        
        struct TestModel {
            let name = "Testing"
        }
        
        let container = Container()
        
        // The issue of read-write access happens due to for example
        // when using the synchronized container in one thread and the original container on another
        // The synchronized container will call syncIfEnabled, blocking anyone that tries to get the mutex
        // But since the method syncIfEnabled only verifies if the mutex is available if the container is marked as
        // synchronized, if the other thread calls the original container then the synchronized flag is false and then
        // the syncIfEnabled ignores the mutex proceeding with the execution.
        
        // This happens due to the synchronyzed container calling parent?.getEntry, accessing the dictionary of the  original container if the service isn't found in the new synchronized container(which will happen almost 100% of the time since we usually add the services to the original container)
        
        // This could also happen when resolving if you use a synchronized container in one thread and the original container in another
        
        // In our case we have some registrations that are done at a later moment, only when needed. Since some functionalities in the app aren't used very often, registering them all at the beginning would only be a burden to the app startup.
        
        // We also don't use a synchronyzed container by default when registering since this isn't recommended
        // and this would probably degrade our performance since it would wait the mutex to execute the whole
        // register/resolve function instead of waiting the mutex only to access the dictionary as in the ThreadSafeDictionary implementation
        
        for _ in 0..<numberOfQueues {
            DispatchQueue.global().async {
                // Only an example to force the crash.
                // This doesn't happen with ThreadSafeDictionary since the implementation would always verify
                // the mutex before accessing the dictionary
                Thread.sleep(until: resolutionDate)
                container.register(TestModel.self) { _ in
                    TestModel()
                }
            }
        }
        
        for _ in 0..<numberOfQueues {
            let expectation = expectation(description: "Should have been resolved")
            expectations.append(expectation)
            DispatchQueue.global().async { [expectation] in
                Thread.sleep(until: resolutionDate)
                _ = container.synchronize().resolve(TestModel.self)
                expectation.fulfill()
            }
        }
        
        wait(for: expectations, timeout: 10)
    }`

maxim-chipeev:
> Hello @maxim-chipeev. I'm facing the same issue and here is a code snippet of a test that is able to replicate the issue. I don't know if their use case is the same as mine but the cause would be the same, an access to the services dictionary from different threads.
> 
> ` func test_SwinjectCrash() { let numberOfQueues = 25 var expectations = XCTestExpectation let resolutionDate = Date() + 1
> 
> ```
>     struct TestModel {
>         let name = "Testing"
>     }
>     
>     let container = Container()
>     
>     // The issue of read-write access happens due to for example
>     // when using the synchronized container in one thread and the original container on another
>     // The synchronized container will call syncIfEnabled, blocking anyone that tries to get the mutex
>     // But since the method syncIfEnabled only verifies if the mutex is available if the container is marked as
>     // synchronized, if the other thread calls the original container then the synchronized flag is false and then
>     // the syncIfEnabled ignores the mutex proceeding with the execution.
>     
>     // This happens due to the synchronyzed container calling parent?.getEntry, accessing the dictionary of the  original container if the service isn't found in the new synchronized container(which will happen almost 100% of the time since we usually add the services to the original container)
>     
>     // This could also happen when resolving if you use a synchronized container in one thread and the original container in another
>     
>     // In our case we have some registrations that are done at a later moment, only when needed. Since some functionalities in the app aren't used very often, registering them all at the beginning would only be a burden to the app startup.
>     
>     // We also don't use a synchronyzed container by default when registering since this isn't recommended
>     // and this would probably degrade our performance since it would wait the mutex to execute the whole
>     // register/resolve function instead of waiting the mutex only to access the dictionary as in the ThreadSafeDictionary implementation
>     
>     for _ in 0..<numberOfQueues {
>         DispatchQueue.global().async {
>             // Only an example to force the crash.
>             // This doesn't happen with ThreadSafeDictionary since the implementation would always verify
>             // the mutex before accessing the dictionary
>             Thread.sleep(until: resolutionDate)
>             container.register(TestModel.self) { _ in
>                 TestModel()
>             }
>         }
>     }
>     
>     for _ in 0..<numberOfQueues {
>         let expectation = expectation(description: "Should have been resolved")
>         expectations.append(expectation)
>         DispatchQueue.global().async { [expectation] in
>             Thread.sleep(until: resolutionDate)
>             _ = container.synchronize().resolve(TestModel.self)
>             expectation.fulfill()
>         }
>     }
>     
>     wait(for: expectations, timeout: 10)
> }
> ```
> 
> `

Hi Nathan, thank you for sharing this snippet—from what I can see that does look unsafe. The Container returned by `.synchronize()` should be the only one referred to by business logic.

What I can do alternatively to this PR, is remove the "synchronization" logic that seems to be confusing—and make Containers inherently threadsafe to access.

Can you try if it crashes with the following being the only synchronize call?
```
let container = Container().synchronize()
```

nathan-batista:
Hello @maxim-chipeev.
I already tried the following approach:

```
let container = Container()
let synchronizedContainer = container.synchronize()

for _ in 0..<numberOfQueues {
    DispatchQueue.global().async {
        Thread.sleep(until: resolutionDate)
        container.register(TestModel.self) { _ in
            TestModel()
        }
    }
}

for _ in 0..<numberOfQueues {
    let expectation = expectation(description: "Should have been resolved")
    expectations.append(expectation)
    DispatchQueue.global().async {
        Thread.sleep(until: resolutionDate)
        _ = synchronizedContainer.resolve(TestModel.self)
        expectation.fulfill()
    }
}
```
But the solution above crashes the app.

The only way to completely solve this issue by using the `synchronize()` function would be to do as below:
```
let container = Container().synchronize() as! Container
for _ in 0..<numberOfQueues {
    DispatchQueue.global().async {
        Thread.sleep(until: resolutionDate)
        container.register(TestModel.self) { _ in
            TestModel()
        }
    }
}
for _ in 0..<numberOfQueues {
    let expectation = expectation(description: "Should have been resolved")
    expectations.append(expectation)
    DispatchQueue.global().async {
        Thread.sleep(until: resolutionDate)
        _ = container.resolve(TestModel.self)
        expectation.fulfill()
    }
}
```
But as i commented in the code snippet i sent previously i believe the performance of this approach would be even worse than using the ThreadSafeDictionary since the whole register and resolve block would be executed by the `syncIfEnabled` function, then each register/resolve call would have to wait the mutex to be available instead of waiting only the access to the dictionary to be available.

maxim-chipeev:
> Hello @maxim-chipeev. I already tried the following approach:
> 
> ```
> let container = Container().synchronize()
> 
> for _ in 0..<numberOfQueues {
>     DispatchQueue.global().async {
>         Thread.sleep(until: resolutionDate)
>         container.register(TestModel.self) { _ in
>             TestModel()
>         }
>     }
> }
> 

Hi Nathan, why not the following?

maxim-chipeev:
> But as i commented in the code snippet i sent previously i believe the performance of this approach would be even worse than using the ThreadSafeDictionary since the whole register and resolve block would be executed by the `syncIfEnabled` function, then each register/resolve call would have to wait the mutex to be available instead of waiting only the access to the dictionary to be available.

I see, what I'm looking to know then is whether registration is something you commonly see in your codebase long after app-launch? In our cases all of our registrations happen at the beginning of the app lifecycle.

nathan-batista:
> > But as i commented in the code snippet i sent previously i believe the performance of this approach would be even worse than using the ThreadSafeDictionary since the whole register and resolve block would be executed by the `syncIfEnabled` function, then each register/resolve call would have to wait the mutex to be available instead of waiting only the access to the dictionary to be available.
> 
> I see, what I'm looking to know then is whether registration is something you commonly see in your codebase long after app-launch? In our cases all of our registrations happen at the beginning of the app lifecycle.

In my use case since i have different modules we only register the essential and the most used features in the beginning of the app life cycle.
Services from modules that aren't used very often are only registered when needed. 
Let's say i have the ModuleA, ModuleB and ModuleC in which ModuleA and ModuleB are essential for my app but ModuleC isn't used very often, in this case i register ModuleA and ModuleB at the beginning of the app life cycle but i only register the ModuleC dependencies when the user starts a journey where this module is needed.

This could also be the case for apps which have features enabled by feature flags.

ThomasDutartre:
Are you waiting something else for this pull request ? @maxim-chipeev

0xLucasMarcal:
@Sali0m added your changes to this PR: https://github.com/Swinject/Swinject/pull/582. Works fine with the latest language mode

PR #582, opened 2025-05-18
Enable Swift 6 language mode

0xLucasMarcal:
# Enable Swift 6 Language Mode

This PR updates Swinject to fully embrace Swift 6.0 and modern Swift concurrency patterns. The changes include both tooling updates and code modernization to leverage Swift 6's features.

It still requires some more profound changes if we want to adopt Swift Concurrency fully, but it seems the case since we want to prioritize performance over using the latest features. In the majority of cases, using an actor or Mutex from the Synchronization framework is twice as slow as using the traditional lock/Fair Lock mechanisms.

### Tooling Updates
- Updated Swift tools version to 6.0.0
- Updated SwiftFormat configuration to target Swift 6.0

### Code Modernization
- Replaced custom logging system with environment variable-based logging (`SWINJECT_LOGGING_ENABLED`)
- Enhanced thread safety by adopting modern Swift concurrency:
  - Converted `ThreadSafeDictionary` from lock-based to actor-based implementation
  - Added `Sendable` conformance to key types
- Improved code organization:
  - Consolidated `ObjectScope` definitions into a single file
  - Removed redundant `ReadWriteLock` implementation in favor of actor-based concurrency
  - Made `ObjectScope` a `final` class and marked it as `@unchecked Sendable`

## Migration Notes
- If you were using the custom logging function, switch to using the `SWINJECT_LOGGING_ENABLED=1` environment variable
- Code using `ThreadSafeDictionary` may need to be updated to handle the actor-based implementation
- The changes maintain API compatibility while improving thread safety and performance

## Testing
Please ensure to test thoroughly, especially in concurrent scenarios, as this update includes significant changes to thread safety mechanisms.


0xLucasMarcal:
cc @maxim-chipeev

maxim-chipeev (approved):

dwrss:
@0xLucasMarcal @maxim-chipeev
This was not an API-compatible change.
Removing `loggingFunction` breaks the builds of anyone using it and the entry in the release notes "Enable Swift 6 language mode" doesn't imply this change. It's also [still documented](Enable Swift 6 language mode).
The `SWINJECT_LOGGING_ENABLED` env variable also is not an equivalent to being able to inject a logging function that puts entries into our own logging system.

